PR #25, opened 2011-08-05
Node & npm

hugowetterberg:
I've modified html-minifier to work with node and npm, without affecting the way it's run in the browser.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,35 @@
+{
+  "name": "html-minifier",
+  "version": "0.43",
+  "description": "HTMLMinifier is a Javascript-based HTML minifier with lint-like capabilities.",
+  "keywords": ["html", "minifier", "lint"],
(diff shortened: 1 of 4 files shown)


kangax:
Thanks Hugo! I've been meaning to do this for a while, but could never find time :)

The only thing... shouldn't htmlparser.js be added as a dependency for htmlminifier.js?

hugowetterberg:
Well, it is included if global doesn't contain HTMLParser (not running in browser) and if require() is available (running in node): 
https://github.com/hugowetterberg/html-minifier/commit/cf25eaa25f8307a129625136d95879c41d76cd20#L2R23

...or did you think of some other sort of dependency declaration?

kangax:
Ah, yes, makes sense. Merged. Thanks again.

PR #27, opened 2012-02-29
customizable canTrimWhitespace & canCollapseWhitespace

crcn:
A good use case is minifying inline templates. For example:

``` javascript
function canRemoveWhitespace(tag, attrs) {

    if(tag != 'script') return false;

    for (var i = attrs.length; i--;) {

        var attr = attrs[i];

        if (attr.name == 'type' && /text\/x-tmpl-\w+/.test(attr.value)) {
            return true;
        } 
    }

    return false;
}

var minified = minify(content,  {
    collapseWhitespace: true,
    canTrimWhitespace: canRemoveWhitespace,
    canCollapseWhitespace: canRemoveWhitespace
});
```

Could minify the `text/x-tmpl-mu` script here:

``` html
<html>
    <head>
        <script type="text/x-tmpl-mu">
            Fancy mustache template:
            <h4>{{title}}</h4>
        </script>
        </head>
        <body>
        </body>
</html>
```

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -251,33 +251,58 @@
     
     return (' ' + attrFragment);
   }
+
+
+  function setDefaultTesters(options) {
+
+    var defaultTesters = ['canCollapseWhitespace','canTrimWhitespace'];
+
+    for (var i = 0, len = defaultTesters.length; i < len; i++) {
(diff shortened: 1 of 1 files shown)


kangax:
Thanks. Looks useful indeed. Would be great to have some unit tests for this too. I'm merging for now.

PR #29, opened 2012-05-10
Improvements for CommonJS support

stereobooster:


kangax:
Thanks! Added couple of comments.

stereobooster:
made fixes for your comments.
P.S. You can leave notes on `Diff` tab on this page, then comments will stay even if the commits will be replaced.

PR #30, opened 2012-05-22
update qunit

stereobooster:
1. Update qunit to be able run tests at serverside.
2. change window to this. Because it cause error in JScript runtime (stupid JScript).


kangax:
Thanks!

stereobooster:
Just published [ruby gem](/stereobooster/html_minifier) thats wraps your library

PR #34, opened 2013-01-10
Problem with comments in tags

gregberge:
Hello,

There is a problem when there is some comments in tags. I wrote a test and make a pull request to reproduce the problem.

```
<div <!-- HTML COMMENT --> class="one-class"></div>
```


kangax:
This doesn't seem to be valid HTML5 (as per http://html5.validator.nu/)

gregberge:
Ok :) so there is no problem.

PR #38, opened 2013-04-02
IE Bug Fix

willfarrell:
Bug fix related to https://github.com/gruntjs/grunt-contrib-htmlmin/issues/13#issuecomment-15769874

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -41,7 +41,7 @@
   }
   
   function isConditionalComment(text) {
-    return (/\[if[^\]]+\]/).test(text);
+    return ((/\[if[^\]]+\]/).test(text) || (/\s*(<!\[endif\])$/).test(text));
   }
   
   function isEventAttribute(attrName) {


kangax:
Thanks. This reminds me that we need to hook up our testsuite with travis...

willfarrell:
Seeing how this repo is used by `grunt-contrib-htmlmin` which is used by every developer I know ... yeah.

kangax:
Ok, just did this. What better way to spend lunch break!

So `npm test` now runs all tests on node (all pass as of now), and I pushed new 0.5.0 version to npm.

willfarrell:
I just had a thought to combine the regex, turns out doing so would be a lot slower (at least in this case). Something for everyone to keep in mind.

http://jsperf.com/html-minifier-regex

kangax:
Also added travis hook, but don't see any builds yet https://travis-ci.org/#!/kangax/html-minifier

willfarrell:
Awesome. You should push an update to https://github.com/gruntjs/grunt-contrib-htmlmin/blob/master/package.json when npm makes the new version live. The commit will have more weight coming from you then myself.

PR #44, opened 2013-04-10
Fix prefix in element endTag

samaxes:
Fixed element's endTags the same way it was fixed for startTags: sohira@2542791


aurhe:
When trying to minify:

``` html
<pg-notification></pg-notification><div></div>
```

it was changing the order of the tags:

``` html
<pg-notification><div></div></pg-notification>
```

This commit fixes it.

samaxes:
This issue is also fixed by #50. Feel free to close this PR.

Issue #46, opened 2013-04-11
Hyphenated tag names producing incorrect output


stephenplusplus:
From: https://github.com/yeoman/yeoman/issues/1010

It seems using a custom hyphenated tag name will return strange markup.

I ran a test:

```
Actual value:
<custom -directive="">Testing 1, 2, 3.</custom>
Expected value:
<custom-directive>Testing 1, 2, 3.</custom-directive>
```

kangax:
This is invalid HTML so [parser](https://github.com/kangax/html-minifier/blob/gh-pages/src/htmlparser.js) doesn't know how to handle it. Need to add support for custom directives to make this work. Perhaps API of some sort to register tag name. Feel free to experiment.

stephenplusplus:
Ah, I see. I wasn't sure if custom markup was intended to be supported. And it added to my confusion when running the test in the browser, the output was fine, but from Node, it failed. With Angular gaining momentum, and web components coming, there are an increasing number of devs who would stand to gain from html-minifier supporting these custom elements. I'll see if I have some time this weekend to play around!

grEvenX:
Fix proposed in https://github.com/kangax/html-minifier/pull/50

kangax:
This is now fixed

PR #48, opened 2013-04-14
HTML5 flow elements, as well as a re-built all.js

stephenplusplus:
Hey,

I took a stab at a couple things real quick. Not sure if this is "master-ready," but it didn't break any tests :+1: 

First of all, I was stumped why the tests I was running through the terminal were throwing errors that the browser test runner weren't. I tracked that down to `dist/all.js`, which seems to have been out of date with the source files. I didn't see a build process for generating/concatenating all.js, so I just copy and pasted in its parts, and voila, it worked.

Once I got that figured out, I created a new map in the parser for [HTML5 flow elements](https://developer.mozilla.org/en-US/docs/HTML/Content_categories#Flow_content), and updated the code to not pressure these elements to have only one type of content within them.

I added a couple tests for these flow elements, as well as a test for custom components, which seem to be working now that all.js is updated.

Let me know what you think. I'm happy to make adjustments or discuss further the best way to proceed with supporting these new flow elements and custom components.

[#46, https://github.com/yeoman/yeoman/issues/1010, https://github.com/yeoman/grunt-usemin/issues/87]

### DIFF
--- a/dist/all.js
+++ b/dist/all.js
@@ -27,7 +27,7 @@
 (function(global){
 
   // Regular Expressions for parsing tags and attributes
-  var startTag = /^<(\w+)((?:\s*[\w:-]+(?:\s*=\s*(?:(?:"[^"]*")|(?:'[^']*')|[^>\s]+))?)*)\s*(\/?)>/,
+  var startTag = /^<([\w:-]+)((?:\s*[\w:-]+(?:\s*=\s*(?:(?:"[^"]*")|(?:'[^']*')|[^>\s]+))?)*)\s*(\/?)>/,
       endTag = /^<\/(\w+)[^>]*>/,
       attr = /([\w:-]+)(?:\s*=\s*(?:(?:"((?:\\.|[^"])*)")|(?:'((?:\\.|[^'])*)')|([^>\s]+)))?/g,
       doctype = /^<!DOCTYPE [^>]+>/i;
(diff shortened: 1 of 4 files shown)


kangax:
@stephenplusplus Looks OK! I see you cleaned up some trailing spaces in parser. That's good. I must have touched that file a while ago, before my editor was configured to strip them automatically :)

Introduction of flow elements makes sense. The only thing I wanted to do is have a way to adhere to HTML5 and HTML4 rules separately. So that folks who want to keep their documents under HTML4 rules could still do that.

I also see that you added ":" and "-" in startTag regex. I suppose that's a reasonable and innocent change. But shouldn't it also be added to endTag?

stephenplusplus:
Hmm, those regex changes must have been introduced elsewhere. `dist/all.js` was out of date from the actual source files that are within it. When I did a copy-paste of the 3 individual files that comprise 'all.js', that regex change came along with it. A blame should show us where that came from. But, the current end tag regex seems to match any closing tag correctly already.

If you'd like to keep HTML 4 and 5 separate, would you prefer 5 be default, and 4 as an option? I'm happy to go back in and make those changes!

kangax:
Probably would keep it optional for now. We can make it into one of the options (such as `collapseWhitespace`, `removeCommentsFromCDATA`, etc.)

stephenplusplus:
Cool. I've got a move ahead of me so my time will be limited, but I'll definitely do this as quickly as I can :)

stephenplusplus:
Only had a bit of a chance to add this as an option, but gave it a shot.

kangax:
@stephenplusplus I merged other pull requests that change parser to behave like HTML5 by default. I would still want to have this option, as in your implementation. Closing this pull request as it's already outdated. Please submit a new one when you have a chance! Thanks.

PR #49, opened 2013-04-15
Allow hyphens to work in tag names

slundberg:
Minifying the following markup results in an incorrect parse with the div ending up inside the custom tag. This causes problems when minifying angular templates since they often use hyphens in custom tag names. I just updated the endTag regExp to include hyphens (the start tag already accepted them).

``` html
<!doctype html>
<html>
<head>
  <title>custom tags dash problem</title>
</head>
<body>
  <customout>
    <custom-in>
    </custom-in>
    <div class="">
      My parent div is a sibling of customin
    </div>
  </customout>
</body>
</html>
```


slundberg:
I should have checked the pull requests....looks

slundberg:
I should have looked at the current pull requests and seen that this is a duplicate from last week. So I am closing this one.

Issue #54, opened 2013-04-25
wrong HTML element nesting after minification


mbarchein:
Found a bug when minifying HTML with `<button>` inside a `<span>`

Input:

``` html
<span>
     <button>
         Hit me
     </button>
</span>
```

Output produced by html-minifier:

``` html
<span></span><button>Hit me</button>
```

Should be:

``` html
<span><button>Hit me</button></span>
```

Thanks.

munsellj:
+1.  
Another Example:

```
<a>
  <p>Link</p>
</a>
```

Becomes:

```
<a></a><p>Link</p>
```

Instead of:

```
<a><p>Link</p></a>
```

I was assuming this was due to collapseWhitespace, but the same issue also occurs when only running removeComments.

serrynaimo:
+1

a-ursino:
+1

josteing:
+1

chrisdhanaraj:
+1, @Taritsyn fix works.

PR #55, opened 2013-04-30
Fix for bug #54 "Wrong HTML element nesting after minification"

Taritsyn:


kangax:
Fixed manually. Thanks!

PR #59, opened 2013-05-14
HTML5 as an option

stephenplusplus:
HTML5 is still the default, however `{ html5: false }` will essentially revert to the old system. Let me know if you would have implemented this differently, and I'll be happy to make adjustments.


kangax:
Looks good, thanks!

PR #67, opened 2013-06-13
Add code, kbd and q to list of tags with space around them.

sndrs:
Currently, these tags get cinched up, e.g. `<p>Here's some<code>code</code>I wrote.</p>` and `<p>I said<q>yes!</q>and left.</p>`.

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -42,7 +42,7 @@
 
   function collapseWhitespaceSmart(str, prevTag, nextTag) {
     // array of tags that will maintain a single space outside of them
-    var tags = ['a', 'b', 'big', 'button', 'code', 'em', 'font','i',  'img', 'kbd', 'mark', 'q', 's', 'small', 'span', 'strike', 'strong', 'sub', 'sup', 'tt', 'u'];
+    var tags = ['a', 'b', 'big', 'button', 'code', 'em', 'font', 'i', 'img', 'kbd', 'mark', 'q', 's', 'small', 'span', 'strike', 'strong', 'sub', 'sup', 'tt', 'u'];
 
     if (prevTag && (prevTag.substr(0,1) !== '/'
       || ( prevTag.substr(0,1) === '/' && tags.indexOf(prevTag.substr(1)) === -1))) {
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -44,11 +44,13 @@
     // array of tags that will maintain a single space outside of them
     var tags = ['a', 'b', 'big', 'button', 'code', 'em', 'font', 'i', 'kbd', 'mark', 'q', 's', 'small', 'span', 'strike', 'strong', 'sub', 'sup', 'tt', 'u'];
 
-    if (prevTag && prevTag !== 'img' && (prevTag.substr(0,1) !== '/' || tags.indexOf(prevTag.substr(1)) === -1)) {
+    if (prevTag && prevTag !== 'img' && (prevTag.substr(0,1) !== '/'
+      || ( prevTag.substr(0,1) === '/' && tags.indexOf(prevTag.substr(1)) === -1))) {
       str = str.replace(/^\s+/, '');
     }
 
-    if (nextTag && nextTag !== 'img' && (nextTag.substr(0,1) === '/' || tags.indexOf(nextTag) === -1)) {
+    if (nextTag && nextTag !== 'img' && (nextTag.substr(0,1) === '/'
+      || ( nextTag.substr(0,1) !== '/' && tags.indexOf(nextTag) === -1))) {
       str = str.replace(/\s+$/, '');
     }
 
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -42,7 +42,7 @@
 
   function collapseWhitespaceSmart(str, prevTag, nextTag) {
     // array of tags that will maintain a single space outside of them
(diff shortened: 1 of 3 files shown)


kangax:
Thanks! Can you please add unit tests for those?

sparanoid:
Wow, @sndrs we are talking about the same thing at the same time: #68

sndrs:
@sparanoid nice! 

@kangax added these tests - but what i don't understand is why removing tags from the collapseWhitespaceSmart() tags array and rerunning doesn't cause the tests to fail...?

sndrs:
sorry - 08aa320 wasn't right. but the tests still don't fail if the tags array is changed e.g. remove a tag from it...

kangax:
@sndrs Are you building dist/*.js? Take a look at https://github.com/kangax/html-minifier/blob/gh-pages/package.json#L32

sndrs:
@kangax i wasn't - you're right. running that jshint command still passes the test, even if the tag is ommited from the minifier source, but included in the test. 

if it's omitted from the minify(), is it just parsed as just text in the test? so that `<q>blah</q>` is parsed like `<qwe>blah</qwe>` would be?

kangax:
@sndrs Are you talking about `jshint ...` passing or `node test.js`?

sndrs:
@kangax I'm sorry - yeah that wasn't clear! Ok:
- exclude `q` from `tags` array in `collapseWhitespaceSmart()` (i.e. `[...'mark', 's',...]`) but include in the test (i.e. `[...'mark', 'q', 's',...]`).
- `jshint src/htmlminifier.js src/htmllint.js` passes.
- `cat src/htmlparser.js src/htmlminifier.js src/htmllint.js > dist/all.js`, no problems.
- `node test.js` passes i.e. `<p>foo <q>baz</q> bar</p>` passes through in the test as  `<p>foo <q>baz</q> bar</p>`  just fine. 

but where i've been using this with grunt, `blah <q>hi</q> blah` was being cinched to `blah<q>hi</q>blah`, and making this same change (adding `q` to the `tags` array) manually in `dist/all.js` in `grunt-contrib-htmlmin` fixed it. 

does that make sense? i must have misunderstood the tests, but they seem right to me?

sndrs:
@kangax ok - sussed the tests: I wasn't passing the `collapseWhitespace` option. 

however, it revealed an issue with non-closing tags like `img` though, so i've made a change for that...let me know if that's beyond the scope of this pull request and i'll remove it.

`input` would also apply here too i'd say? should be treated like `img`?

sparanoid:
Just fyi I found a list of HTML element maintained by Mozilla:

https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/HTML5/HTML5_element_list#Text-level_semantics

I think all text-level semantics shouldn't be collapsed, But I don't know if there's any side effect I haven't test them yet.

sndrs:
@sparanoid that seems a good list to work with, but `<br>` and `<wbr>` for example don't seem like they would need to maintain space? let's see if @kangax has any thoughts.

currently, the minifier list omits the following tags that are present in the moz list:
- `<cite>`
- `<dfn>`
- `<abbr>`
- `<data>`
- `<time>`
- `<var>`
- `<samp>`
- `<ruby>`, `<rt>`, `<rp>`
- `<bdi>`
- `<bdo>`

which all seem like candidates to me, and then
- `<br>` 
- `<wbr>`

which don't.

src/htmlminifier.js
```diff
@@ -72,7 +72,7 @@
 
   function canRemoveAttributeQuotes(value) {
     // http://mathiasbynens.be/notes/unquoted-attribute-values
-    return /^[^\x20\t\n\f\r"'`=<>]+$/.test(value);
+    return (/^[^\x20\t\n\f\r"'`=<>]+$/).test(value);
```
sndrs, line 75: This was just needed for linting.

kangax:
Thanks for the tests and investigation, @sndrs!

FWIW, we already store a [list of inline elements in parser](https://github.com/kangax/html-minifier/blob/gh-pages/src/htmlparser.js#L42). Should just expose it from there. It's based on HTML4.01, so might need an update.

As far as img tag — why exactly do we need a check for it? Is it because it's an empty element? In that case, we could also utilize [list of empty elements](https://github.com/kangax/html-minifier/blob/gh-pages/src/htmlparser.js#L36) from parser.

sndrs:
Yeah I wondered about the lists in parser. I'm not sure it's just whether an element is inline, though, that's the issue? For example, a `script` tag wouldn't need any space around it. I'd expect a minifier to remove all white space there I think. Similarly `basefont`, for example. I feel like maintaining a manual list is the best move here, based on what you'd expect to happen during minification. 

I think this pertains to images (and some other empty elements) too. I totally agree a list of empty elements is better option than just a quick check for an image. But as with `basefont`, I wouldn't expect `base` to preserve a maximum of 1 space around it, but removing  _all_ space from around an `img` could affect layout, which I don't think I'd expect e.g. if

```
<p>Here is a tiny little image <img height="11" width="11 /> I made.</p>
```

became

```
<p>Here is a tiny little image<img height="11" width="11 />I made.</p>
```

I don't think we should rely on CSS to ensure space around an in-text `img` - the HTML should work without styles. 

That said, something like

```
<p>Here a tiny little image 
    <img height="11" width="11 /> 
I made.</p>
```

should definitely be condensed to 

```
<p>here a tiny little image <img height="11" width="11 /> I made.</p>
```

imo.

Does that make sense?

kangax:
@sndrs Makes sense. I'll merge this pull request for now, and we can consider more complete support in the future.

PR #68, opened 2013-06-13
Avoid collapsing `code` inline tag

sparanoid:
I think there's a bug in current version:

``` javascript
    var input = 'a <code>hello world</code> example.';
    var output = minify(input, { collapseWhitespace: true });

    output; // 'a<code>hello world</code>example.'
```

`code` is an inline element, collapsing should be avoided.


sparanoid:
Dup: #67

PR #93, opened 2013-10-22
Preserve quotes if attribute ends with a trailing slash

stieler-it:
minified URLs in parameters, like `<a href="http://www.example.com/">` need to keept their quotes because `<a href=http://www.example.com/>` is interpreted as an URL without trailing slash

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -76,7 +76,9 @@
 
   function canRemoveAttributeQuotes(value) {
     // http://mathiasbynens.be/notes/unquoted-attribute-values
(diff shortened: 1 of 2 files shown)


kangax:
Thanks! Can you please add some unit tests and make sure no existing ones break?

stieler-it:
As requested I added a unit test that fails in current implementation and works with the fix. However, I had to change an already existing unit test: This test was correct and actually it would be possible to strip quotes if the attribute containing a trailing slash is not the last one. I assume that not breaking semantics has higher priority than saving some bytes, though.

kangax:
@stieler-it Definitely! Merging. Thanks again.

PR #94, opened 2013-11-04
Newlines should be collapsed to single space.

rolftimmermans:
Currently a single newline is completely removed if whitespace collapsing is enabled. This is not correct. It is treated as whitespace by browsers and should be collapsed to a single space.

Pull request includes fix and test case.


kangax:
Thanks!

PR #108, opened 2014-01-22
Various improvements

XhmikosR:
@kangax:Let me know if you want me to change anything and I will rebase the branch.


XhmikosR:
http://xhmikosr.github.io/html-minifier/
http://xhmikosr.github.io/html-minifier/tests/

kangax:
@XhmikosR Everything looks good, thanks! The only question — why the removal of some stuff from .jshintrc? Are these the declarations that are being dropped in 3.0?

XhmikosR:
@kangax: I basically removed most options that are on by default. And sorted the remaining entries by name.

BTW, Travis doesn't seem to show for the PR, would be nice to enable it.

XhmikosR:
Another thing, what do you think about moving the web related files to a separate folder? Like "web"? We can make Jekyll use that folder for page generation and the source tree should be cleaner.

Edit: Actually this doesn't work because then the dist directory isn't copied in the output dir.

kangax:
@XhmikosR not sure what's up with Travis; I just enabled it in the profile (had .travis.yml in repo but forgot to enable it on travis side) but it's not appearing yet. Maybe needs more time or maybe something else is off.

XhmikosR:
Maybe it's because the main branch is gh-pages?

kangax:
@XhmikosR maybe... removed master branch; should have done it a while ago : )

PR #110, opened 2014-01-23
Travis

XhmikosR:
Now Travis should work hopefully.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,7 @@ node_js:
   - "0.8"
   - "0.10"
   - "0.11"
+
+branches:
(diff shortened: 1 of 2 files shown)


XhmikosR:
https://travis-ci.org/kangax/html-minifier/builds/17493755

kangax:
Awesome, thanks!

XhmikosR:
No problem, every day we learn something new :)

BTW, do you really want to support Node.js v0.6? If not you should remove it from travis and bump `engines` in package.json.

kangax:
Should be all good now

PR #114, opened 2014-01-28
Switch to Grunt

XhmikosR:
TODO:
- [x] Add banner in the minified file
- [x] Drop Node.js 0.6 support since 0.8 is required by Grunt.
- [x] Enable jshint for all the files
- [ ] Use grunt-contrib-qunit for tests
- [ ] Update README.md with more info about Grunt


XhmikosR:
@kangax: any thoughts so far?

I'm having issues setting up `grunt-contrib-qunit` and finding a solution for the source files directory structure. But one thing at a time :)

kangax:
Damn, everyone is dropping poor node 0.6 these days. Couple months ago, I did a poll and some folks still using it — https://twitter.com/kangax/status/396317222064648193

Other than that, looks good. I'm not familiar with Grunt really. I like the "dist/all.js → dist/htmlminifier.js" change (should have done a while ago, duh).

Thanks!

XhmikosR:
@kangax: rebased against the latest changes. I'm not sure I can fix the remaining issues though. If you are not in a hurry I can look closer the next days.

kangax:
@XhmikosR I don't think we're in a hurry :) Thanks for doing this.

XhmikosR:
@kangax: Almost done. Opinions so far?

Not sure how to make grunt-contrib-qunit to work yet that's why the PR is stalled.

Also about the license headers, I'm thinking maybe moving the htmlminifier.js's header as the main project's header.  What do you think about that?

XhmikosR:
Alternatively I can try using grunt-exec to run `node test.js` until we have a better solution.

XhmikosR:
I think this is ready to be merged. If someone else can make use of another more appropriate plugin for the tests, that would be perfect. Otherwise, things work fine as they are too :)

Not sure what to write in the README about Grunt though.

XhmikosR:
Another thing while I'm at it, what do you prefer, single or double quotes?

kangax:
> Another thing while I'm at it, what do you prefer, single or double quotes?

Single.

Thanks again! Merging.

PR #132, opened 2014-03-08
Tweak JSHint options and add JSCS support

XhmikosR:
Let me know what you think about the JSHint option changes.

BTW, I would enable `curly` and `strict` and remove `laxbreak` and `loopfunc`. IF you agree with those too let me know and I'll update the PR.


kangax:
I agree with `newcap` but usually don't care for `forin` (more on that [here](https://github.com/kangax/fabric.js/issues/1160#issuecomment-35442737)). `curly` is a must, I'm surprised it wasn't enabled. Does `strict` require us to add "use strict" or just enforces strict-mode rules implicitly?

XhmikosR:
OK, I will make those changes tomorrow.

BTW, even grunt-contrib-uglify doesn't support node.js 0.8 anymore it seems...

As for strict, it requires to use the `"use strict";` part.

XhmikosR:
I went ahead and made the changes. How do you want to handle node.js 0.8? Personally, I believe it's fine to drop 0.8 if you bump the version to like 0.6.0.

kangax:
I wouldn't want to drop 0.8 just yet. Seems premature. I was hesitant to remove 0.6 from Fabric just few months ago, since some of the users were still on it. I imagine quite few are still on 0.8.

XhmikosR:
Well, the core code works fine; the plugins for development don't work with  < 0.10...

XhmikosR:
I went ahead and added JSCS using the config from fabric.js. I made a couple of changes there, but everything should be ok.

Please review and let me know if there are any issues :)

XhmikosR:
Something does not work with node.js 0.8 it seems :/

PR #135, opened 2014-03-12
Downgrade qunit so that node.js 0.8 works again.

XhmikosR:
More and more issues arise from supporting node.js 0.8...

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,9 +44,9 @@
     "grunt-contrib-qunit": "0.4.x",
     "grunt-contrib-uglify": "0.3.x",
     "grunt-exec": "0.4.x",
-    "grunt-jscs-checker": "~0.4.0",
+    "grunt-jscs-checker": "0.4.x",
     "load-grunt-tasks": "0.3.x",
-    "qunit": "0.6.x",
+    "qunit": "0.5.x",
     "time-grunt": "0.2.x"
   },
   "files": [


kangax:
: /

XhmikosR:
BTW, maybe you could make a new npm release? There aren't any important changes so far, but it shouldn't hurt either :)

kangax:
Sure. Pushed, published.

XhmikosR:
Thanks!

kangax:
Btw, I just created 0.6.0 milestone and added some issues in there that I'd
like to do next --
https://github.com/kangax/html-minifier/issues?milestone=1&page=1&state=open

I'm pretty swamped with work at the moment, and there's also Fabric issues
piling up. I'll try to get to some of these in a near future. Shooting for
1-2 month to finish and push 0.6.0.

XhmikosR:
Sounds good! I guess you can update the deps in that branch and drop node.js 0.8 support since it's becoming a problem more every day...

kangax:
Yeah, we should probably drop it together with 0.6.0 release.

XhmikosR:
BTW, personally apart from the ability to minify inline javascript, I miss the opportunity to cleanly ignore specific code. I use `<!--googleoff: all-->` etc and I only need to keep that from the comments but it's so much trouble to work around this :/

kangax:
Ok, added.

PR #137, opened 2014-03-18
Minor tweaks

XhmikosR:
BTW I get 4 failed tests in script and style minification.

```
Mozilla/5.0 (Windows NT 6.1; rv:29.0) Gecko/20100101 Firefox/29.0

Tests completed in 106 milliseconds.
366 assertions of 370 passed, 4 failed.
```

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -1,3 +1,5 @@
+/* global CSSOCompressor: true, CSSOTranslator: true, cleanInfo: true, srcToCSSP: true */
+
 (function(global) {
   'use strict';
 
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,10 +1,4 @@
 {
-  "globals":    {
-    "CSSOCompressor": true,
-    "CSSOTranslator": true,
-    "cleanInfo": true,
-    "srcToCSSP": true
-  },
   "browser":    true,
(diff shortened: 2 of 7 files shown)


src/htmlminifier.js
```diff
@@ -1,3 +1,5 @@
+/* global CSSOCompressor: true, CSSOTranslator: true, cleanInfo: true, srcToCSSP: true */
```
kangax, line 1: Hm... yeah, it does make sense to localize these. Thanks.

kangax:
> 366 assertions of 370 passed, 4 failed.

Ahha... I referenced minifier files on a test page and it's (almost) good now. We should really package js/css minifiers together with minifier. We'd probably need to make it modular so that you could specify whether you want them included in a build.

XhmikosR:
Now only one test fails :)

kangax:
Yep, it's the difference in "o" vs. "n" for the mangled variable : )

XhmikosR:
Maybe disable mangle in uglify for the tests?

kangax:
We could, but I wanted to actually ensure that mangling is happening as that's one of the main things that reduces string size.

XhmikosR:
I suggested to disable it for the tests only if possible. Unless you have another workaround on your mind of course :)

PR #141, opened 2014-03-20
Switch to clean-css

XhmikosR:
The web version doesn't seem to work right. But @GoalSmashers [said](https://github.com/GoalSmashers/clean-css/issues/260#issuecomment-38108494) it works for him.

@kangax: before merging this, please make sure we have a working browser version; I'd hate to go through with this and leave it unfinished :/


kangax:
Ok, so the problem was that you declared `CleanCSS` in a local scope, so it shadowed whatever `CleanCSS` was available outside of the function. I've done this mistake so many times... :) Javascript! https://github.com/kangax/html-minifier/commit/0eebc503f0b37becce329375cf1cdfcb0b99fcd3#diff-e7f7e7412b2bb4403d5579516fc6b46fL742

XhmikosR:
Awesome, thanks!

Maybe we could clean up that code? I mean `return new CleanCSSModule().minify(text);` is the same in both cases.

kangax:
Hm, there's few ways we can juggle with it. Maybe something like this:

``` js
var CleanCSS = global.CleanCSS || require('clean-css');
if (!CleanCSS) return text;
return new CleanCSS().minify(text);
```

(and wrapped in try-catch coz `require` might error out)

or something like:

``` js
var CleanCSS = typeof require == 'function' ? require('clean-css') : global.CleanCSS;
if (!CleanCSS) return text;
return new CleanCSS().minify(text);
```

XhmikosR:
It seems there are cases minification doesn't happen for CSS. I mean I see the spaces after semicolons in `style=""` not being removed.

Example

``` html
<td colspan="2" style="text-align:center;padding:0; line-height:1.2em;;">
```

becomes

``` html
<td colspan=2 style="text-align:center;padding:0; line-height:1.2em;">
```

XhmikosR:
I'll leave the refactoring to you :)

BTW, I opened a PR for the cleancss browserify version. Perhaps we could just create it ourselves in the future. But all this after the assets refactoring :)

kangax:
It's good that mention style attributes coz I don't think we pass them through minifyCSS at all...

kangax:
Similarly with minifyJS and on\* attributes

XhmikosR:
OK, I noticed it and thought I'd bring it up :)

Another thing I was thinking. Maybe we should pay it on the safe side and disable clean-css's advanced optimizations by default. Also, I think there should be a way to pass options to clean-css like `compatibility`. What do you think?

kangax:
Sure. I think that's a good idea.

Issue #9, opened 2010-05-17
Remove as many empty/blank attributes as possible (not just core ones)


kangax:

duncanbeevers:
What do you intend for this to do?

Would you like to omit all valueless, non-boolean attributes?

``` html
var input = '<div data-foo></div>';
expect(minifiy(input, { removeEmptyAttributes: true }), '<div></div>');
```

This seems like a reasonable behavior, although I haven't seen any markup that would benefit from it, and I have seen markup that would be damaged by it, such as angular templates driven by the presence of some data attribute.

Can you think of any consequences to this?

kangax:
I don't remember exactly, but I think I wanted to clean up cases like this:

``` html
<p foo=" "></p>
<p foo=""></p>
<p foo></p>
```

 As of now, I'm inclined to just close this until we get some real, valid use cases.

Sounds reasonable?

duncanbeevers:
Agreed.

ghost:
Turns out pre-rendering a Vue app creates lots of markup in the fashion of `<div v-12345678="">`, wasting an extra three bytes per element. Doesn't look like `removeEmptyAttributes` covers this yet. I have around 250 of those per document.

Is there any chance we could get this issue re-opened?

DamonHD:
Sure that some combination of:

```
          --remove-attribute-quotes 
          --remove-redundant-attributes
          --remove-empty-attributes

```
doesn't do what you want already?

FWIW, my script to drive html-minifier is at:

    http://earth.org.uk/script/html5-minify.sh

Rgds

Damon

ghost:
Yeah, I'm sure. I've tested enabling all three of them simultaneously before commenting. Test case:

`<div v-12345678=""></div>`

which stays the same.

PR #148, opened 2014-03-25
Don't add empty string value to valueless attributes

remcohaszing:
For example does not change `<div ng-transclude></div>` to `<div ng-transclude=""></div>`, but leaves empty attributes as is. Possibly fixes issue #9, although I am not sure what is meant there exactly.


kangax:
Thanks!

Issue #169, opened 2014-05-06
Prevent TypeError on empty values of certain attributes


janeklb:
Certain attributes (eg. `href`) will get their values trimmed, even if they're not set - which throws: `TypeError: Cannot call method 'trim' of undefined`

The trim function could do a type check to prevent this.

(similar to #150)

PR #170, opened 2014-05-06
checking for string type before calling trim() (for issue #169)

janeklb:
Proposed fix for https://github.com/kangax/html-minifier/issues/169


kangax:
Thanks!

PR #184, opened 2014-06-03
fix #178

### LINKED ISSUE #178: Minifier removes relevant whitespace in collapseWhitespace mode (e. g., nbsp)
When presented with HTML content like `<span>12 %</span>` (notice that the space between 12 and % is in fact U+00A0, the non-breaking space otherwise known as `&nbsp;`), collapseWhitespace() removes it and replaces it with a normal space (U+0020).

There is quite a [zoo](https://www.cs.tut.fi/~jkorpela/chars/spaces.html) of special whitespace characters which should – imho – not be modified by the minifier.

As far as I can tell, this problem could mostly be solved by using `/[\t\n\r]+/g` as the regular expression in collapseWhitespace().

addaleax:
After looking a bit through the minifier code, it seems this is the only place where #178 can make trouble

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -37,7 +37,7 @@
   }
 
   function collapseWhitespace(str) {
-    return str ? str.replace(/\s+/g, ' ') : str;
+    return str ? str.replace(/[\t\n\r ]+/g, ' ') : str;
   }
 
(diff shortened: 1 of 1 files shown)


kangax:
Thanks!

PR #186, opened 2014-06-12
Add SVG element

jonathantneal:
When given a width and height, SVGs are inline. This is often the case when SVGs are used as icons.


kangax:
Can you please add a unit test as well?

jonathantneal:
Certainly. Let me know if this will do.

kangax:
Thanks!

PR #190, opened 2014-06-18
Add support for <wbr> tag.

osamesama:
The current implementation does not appear to be aware of the <wbr> tag.  This is the "word break opportunity" tag, which is processed as a void element.  Without proper recognition of the <wbr> tag, htmlmin unfortunately adds illegal closing tags when processing HTML files that use it.

See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/wbr

This pull request proposes a possible enhancement to support <wbr>.

### DIFF
--- a/src/htmlparser.js
+++ b/src/htmlparser.js
@@ -41,7 +41,7 @@
       endIgnore = /(%|\?)>/;
 
   // Empty Elements - HTML 4.01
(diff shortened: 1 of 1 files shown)


kangax:
Thanks @osamesama! Can you please add a unit test as well?

osamesama:
You're welcome.  I'll see what I can do.

Issue #172, opened 2014-05-19
Incorrect minification


pallix:
Hello.

The content of this template:

   &lt;div ui-view=''&gt;<br/>     &lt;h1&gt;{{title}}&lt;/h1&gt;<br/>     &lt;p class='introduction' translate='lican.intro.text'&gt;&lt;/p&gt;<br/>      &lt;form ng-submit='startAnalysis()'&gt;<br/>        &lt;button class='btn btn-default' translate='lican.intro.start' type='submit'&gt;    &lt;/button&gt;<br/>      &lt;/form&gt;<br/>    &lt;/div&gt;<br/>

is incorrectly minified to:

```
"<div ui-view><h1>{{title}}</h1><p class=introduction translate></p><form ng-submit=startAnalysis()><button class=\"btn btn-default\" translate type=submit></button></form></div>"
```

The translate attribute loses its value.

I had the bug from grunt html2js and putting some log shows a problem in the minify function. See also https://github.com/karlgoldstein/grunt-html2js/issues/42

kangax:
Looking at this closer, I see that attribute value is being removed because translate is a boolean attribute and so empty value is the same as a "yes" string.

> The attribute's keywords are the empty string, yes, and no. The empty string and the yes keyword map to the yes state. The no keyword maps to the no state. In addition, there is a third state, the inherit state, which is the missing value default (and the invalid value default).

http://www.whatwg.org/specs/web-apps/current-work/multipage/elements.html#the-translate-attribute

kangax:
So the question is: why is there "lican.intro.start" value, if browser only understands "yes", "no", and empty string.

pallix:
It's because of the angular translate plugin. http://angular-translate.github.io/

lggwettmann:
I think i found same problem here https://gist.github.com/lggdance/ac6b052292100c95a2e9 - check the view-checkin.html which's <signup> directive loses the value of the open-attribute from

``` html
<signup ng-show="openit" name="{{input}}" open="openit" guest="guest" on-signup="addGuest()"></signup>
```

to

``` html
<signup ng-show="openit" name="{{input}}" open guest="guest" on-signup="addGuest()"></signup>
```

.

Any solutions to that yet?

kangax:
Well, one way is not to use "collapse boolean attributes" — http://kangax.github.io/html-minifier/

lggwettmann:
Thanks for pointing that out :)

schmod:
[...]

PR #193, opened 2014-06-23
Do not treat translate attribtues as boolean values

schmod:
See #172 and the current draft of the HTML5 spec:
http://www.whatwg.org/specs/web-apps/current-work/multipage/elements.html#the-translate-attribute

Note that the `translate` spec is somewhat confusing, as elements inherit their `translate` value from their parent node(s), which is a fairly unusual behavior.  Thus, `translate` appears to be exposed to the DOM as a boolean, but requires a string value of `"yes"` or `"no"` to be set in the HTML document itself (ergo, it is not a "boolean" attribute from html-minifier's perspective).

Given that no browsers currently implement this API, and a few [libraries](http://angular-translate.github.io) happen to use custom `translate` attributes in the DOM, it seems pretty safe to remove this for now.

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -155,7 +155,7 @@
   }
 
   function isBooleanAttribute(attrName) {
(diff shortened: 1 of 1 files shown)


kangax:
Fair enough. Thanks!

Issue #198, opened 2014-07-02
caseSensitive: true does not work (completely)


JakobJingleheimer:
Tag names are always converted to lowercase irrespective of this flag:

[`L466`](https://github.com/kangax/html-minifier/blob/gh-pages/src/htmlminifier.js#L466) `tag = tag.toLowerCase();`

[`L466`](https://github.com/kangax/html-minifier/blob/gh-pages/src/htmlminifier.js#L466) `buffer.push('</', tag.toLowerCase(), '>');`

JakobJingleheimer:
Changing

[`L466`](https://github.com/kangax/html-minifier/blob/gh-pages/src/htmlminifier.js#L466) `tag = tag.toLowerCase();`
→ `tag = options.caseSensitive ? tag : tag.toLowerCase();`

[`L522`](https://github.com/kangax/html-minifier/blob/gh-pages/src/htmlminifier.js#L522) `buffer.push('</', tag.toLowerCase(), '>');`
→ `buffer.push('</', options.caseSensitive ? tag : tag.toLowerCase(), '>');`

Resolves this.

kangax:
Thanks! Can you please submit a PR (ideally, with a test case)?

JakobJingleheimer:
Yep. I'll do it tonight when i get home from work.

JakobJingleheimer:
Grr. I kid you not—just as I run `git push`, my home internet died.

What testing framework does this use? It was simple enough to update the exisiting tests for caseSensitive, but I'd like to run it just to be sure.

PR #200, opened 2014-07-04
'caseSensitive' option now applies to tags as well as attributes

JakobJingleheimer:
Addresses #198

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -463,7 +463,7 @@
           return;
         }
 
-        tag = tag.toLowerCase();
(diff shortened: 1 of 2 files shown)


kangax:
Great, thanks!

JakobJingleheimer:
How does the `/dist` directory get updated?

kangax:
As part of `grunt` (or only concat task)

JakobJingleheimer:
Oh whoops. Should i run that and re-submit? Also, any chance this could get tagged for release?

kangax:
Yes, please resubmit with already bumped version. I'll publish.

PR #206, opened 2014-07-10
Don’t choke on valueless event attributes

duncanbeevers:
As discovered by @toxinhead in #205, event attributes without values cause minification errors.


kangax:
Nice, thanks!

boycce:
Great work.

PR #207, opened 2014-07-10
Limit line length

duncanbeevers:
Specify maximum line length, split at valid html split-points.


kangax:
Looks good @duncanbeevers, thanks! I can't merge it automatically because I think dist file conflicts with the one I built recently and pushed. Could you please resubmit?

duncanbeevers:
I think in the future I'll leave these `dist` commits off until post-merge.

kangax:
Thanks! Yeah, it's best to omit dist completely. I usually build and publish occasionally.

Issue #166, opened 2014-04-30
<i>, <b>, & <u> are OK now


tomByrer:
I know it was hip to avoid using the `<i>` & `<b>` (& somewhat `<u>` thought that didn't seem as popular), perhaps it was in HTML4 specs to avoid, but [HTML5.1 says it is OK to use `<i>, <b>, & <u>` again](http://www.w3.org/html/wg/drafts/html/master/single-page.html#the-i-element).  & even if not, I'm going to still use them.

Could the `Found presentational <i> element` warnings be turned off please, or at least configurable in the web GUI?  They're just noise, & I end up ignoring all the warnings.  I can try to hunt & peck to in your code to do this, but I wanted OK before I commenced please.

kangax:
Yeah, I suppose we could do that. I don't think people pay much attention to those warnings but still...

tomByrer:
Thanks!
I'm starting to use HTML-minifier as you use [JSCritic](http://jscritic.com/); to quality check foreign HTML.  eg I think `Found style attribute on <div> element` useful to know; I can move that `style="__"` into my .CSS.  All the other 18 `Found presentational ...` I don't care about though.
This will help s/n.

duncanbeevers:
Most font icon libraries use the `<i>` element to denote glyphs. Seems reasonable to remove.

tomByrer:
Thanks!

PR #212, opened 2014-07-28
<s> has been reinstated

tomByrer:
http://www.w3.org/html/wg/drafts/html/master/single-page.html#the-s-element
ref #166


kangax:
Thanks!

PR #214, opened 2014-07-29
Gzip benchmark

rsp:
I added a benchmark for minified and gzipped HTML

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,5 +89,17 @@ var result = minify('<p title="blah" id="moo">foo</p>', {
 result; // '<p title=blah id=moo>foo</p>'
 ```
 
+Running benchmarks
+------------------
+Benchmarks for minified HTML:
+```
+node benchmark.js
+```
+
+Benchmarks for minified and gzipped HTML:
+```
+node gzip-benchmark.js
+```
+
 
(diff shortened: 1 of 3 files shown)


kangax:
Hey @rsp, thanks for this! Looks great except that we shouldn't have 2 ways to run benchmark. We should report gzipped sizes/comparison via main benchmark. This would also allow to avoid all the repetition in gzip-benchmark.js (once I add another file to test, I'll need to update gzip-benchmark.js as well, etc.)

rsp:
Hey, @kangax, I completely agree. I've done it that way because I just wanted to quickly see what the results would be on a gzipped data and that was the way that was sure to not break anything in the process. If you think it'd be useful to have then I can modify the main benchmark to add gzipped tests.

I think it would be most useful to have all of those values in one table so instead of: File, Before, After, Savings, Time there would be File, Before, After, Before gz, After gz, Savings, Savings gz, Time (time only for unzipped because there is no need to count the gzipping time).

Another way would be to have two lines in every table cell - unzipped above, gzipped below. When I have some time I can do it in a more useful way. I wasn't sure anyone would find it useful.

kangax:
Absolutely. I like 2 line solution the most. And again, thanks for digging into this!

kangax:
@rsp would still love to see this integrated into the main benchmark btw! :)

PR #232, opened 2014-08-15
Use native Array#reduce instead of _.reduce

shinnn:
Array.prototype.reduce:
https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

Underscore.js is unnecessary in this case.


kangax:
Thanks!

shinnn:
No problem!

Issue #185, opened 2014-06-04
Invalid minification with img when removeAttributeQuotes=true and keepClosingSlash=true


Appsens:
Hello,
When the 2 options removeAttributeQuotes and keepClosingSlash are activated, the following element `<img src="img/logo.png" />` is minified to `<img src=img/logo.png />`. It looks ok, but the web browsers try to load "img/logo.png/" (with a trailing '/') which causes(most of the time) a 404 error.
In this case (src is the last attribute of an img), the quotes shouldn't be removed.
Best regards.

PR #238, opened 2014-08-20
Quote terminal attribute value for unary slash elements

### LINKED ISSUE #233: self closing tag after attribute not being handled correctly
In other words `<link rel="stylesheet" href="/bundles/links-c90492ab.css"/>` gets minified to `<link rel=stylesheet href=/bundles/links-c90492ab.css/>` which will make a request for `'/bundles/links-c90492ab.css/'` which will 404 due to the trailing slash.

Settings:

```
removeComments: true,
collapseWhitespace: true,
collapseBooleanAttributes: true,
removeAttributeQuotes: false,
keepClosingSlash: false,
removeRedundantAttributes: true,
useShortDoctype: true,
removeEmptyAttributes: true,
removeOptionalTags: true,
removeEmptyElements: false,
keepClosingSlash: true
```

Should remove closing slash before removing attribute quotes.

duncanbeevers:
Fixes #233

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -307,11 +307,12 @@
     return markup;
   }
 
-  function normalizeAttribute(attr, attrs, tag, options) {
+  function normalizeAttribute(attr, attrs, tag, unarySlash, index, options) {
 
     var attrName = options.caseSensitive ? attr.name : attr.name.toLowerCase(),
         attrValue = attr.escaped,
-        attrFragment;
+        attrFragment,
+        isTerminalOfUnarySlash = unarySlash && index === attrs.length - 1;
 
     if ((options.removeRedundantAttributes &&
       isAttributeRedundant(tag, attrName, attrValue, attrs))
@@ -327,7 +328,7 @@
     attrValue = cleanAttributeValue(tag, attrName, attrValue, options, attrs);
 
     if (attrValue !== undefined && !options.removeAttributeQuotes ||
-        !canRemoveAttributeQuotes(attrValue)) {
+        !canRemoveAttributeQuotes(attrValue) || isTerminalOfUnarySlash) {
       attrValue = '"' + attrValue + '"';
     }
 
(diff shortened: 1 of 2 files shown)


src/htmlminifier.js
```diff
@@ -307,11 +307,12 @@
     return markup;
   }
 
-  function normalizeAttribute(attr, attrs, tag, options) {
+  function normalizeAttribute(attr, attrs, tag, unarySlash, index, options) {
```
kangax, line 310: Hm, we should probably extract args into an object at some point, now that it's up to 6

tests/minifier.js
```diff
@@ -697,6 +697,9 @@
 
   test('keeping trailing slashes in tags', function() {
     equal(minify('<img src="test"/>', { keepClosingSlash: true }), '<img src="test"/>');
+    // https://github.com/kangax/html-minifier/issues/233
+    equal(minify('<img src="test"/>', { keepClosingSlash: true, removeAttributeQuotes: true }), '<img src="test"/>');
+    equal(minify('<img title="foo" src="test"/>', { keepClosingSlash: true, removeAttributeQuotes: true }), '<img title=foo src="test"/>');
```
kangax, line 702: If it's `` it will remove quotes from "src", right (since it's not the last attribute)? It could make sense to add test for that particular case.

Also, as far as attribute goes — we're keeping quotes around all of them, not just "src" (based on what I see in your changes), right? Do you think we should perhaps check attribute name or is that a path to more issues in the future?

src/htmlminifier.js
duncanbeevers, line 310: I considered not passing in the index and simply comparing `attr` to `attrs[attrs.length - 1]`, but this felt a little clearer.

tests/minifier.js
duncanbeevers, line 702: In this PR, we retain quotes around the ultimate attribute for any element, regardless of the attribute's name.

For example, if we have an element like this.

```
<img data-foo="bar" />
```

It would be inappropriate to minify it thus

```
<img data-foo=bar/>
```

In this case, the data attribute value would come back as `"bar/"`.

So it's not about quoting based on any specific attribute name, but instead based on attribute position.

Alternatively, we could just emit an extra space before the closing unary slash for elements with attributes

```
<img/>
<img data-foo=bar />
```

Not sure why I didn't think of that before :)

duncanbeevers:
Fixes #185

tests/minifier.js
kangax, line 702: Makes sense!

PR #239, opened 2014-08-20
Gzip benchmark

duncanbeevers:
Based on @rsp's work, and supersedes #214

screen shot 2014-08-20 at 3 11 36 pm


kangax:
Hm, can't seem to merge automatically. Mind resubmitting on top of latest master?

kangax:
Also... 7.15KB gzipped savings on Amazon? Impressive :)

duncanbeevers:
I noticed on `gh-pages`, that just running the benchmarks causes changes to the committed, minified files.

First, I think those assets shouldn't be under source control, and we should just rely on the benchmark to generate them as needed. The fact that they're drifting seems a little odd.

duncanbeevers:
I've removed all the benchmark-generated assets from source control and tidied up the `.gitignore` to make sure they don't accidentally get committed.

PR #241, opened 2014-08-20
Readme cleanup

duncanbeevers:


duncanbeevers:
Removed separate gzip benchmark instructions and BitDeli integration

PR #242, opened 2014-08-21
Case-sensitivity edge-cases

duncanbeevers:
I did a quick audit of the codebase to see if there were areas where oddly-cased tag and attribute names might be mishandled, and found quite a few.

In many cases I was able to simply pass around the lower-cased tag name but in order to maintain case-sensitivity when configured, I also have to schlep around the original format for tag names.

My hope was that reducing the number of string conversions would result in an overall speed-up, but it appears the additional memory pressure means this is pretty much a wash.

### DIFF
--- a/tests/minifier.js
+++ b/tests/minifier.js
@@ -135,6 +135,20 @@
 
     input = '<STYLE><!-- alert(1) --><\/STYLE>';
     equal(minify(input, { removeComments: true }), '<style><!-- alert(1) --><\/style>');
+
+    // Special elements
+    var elements = ['script', 'style', 'noscript'];
+    var element;
+    for (var i = elements.length - 1; i >= 0; i--) {
+      element = elements[i];
+      input = '<' + element + '><!-- alert(1) --></' + element + '>';
+      output = input;
+
+      equal(minify(input, { removeComments: true }), output);
+
+      input = '<' + element[0].toUpperCase() + element.slice(1) + '><!-- alert(1) --></' + element + '>';
+      equal(minify(input, { removeComments: true }), output);
+    }
   });
 
   test('ignoring comments', function() {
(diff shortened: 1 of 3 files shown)


duncanbeevers:
This also deals with one previously-overlooked edge case.

``` html
<div style></div>
```

tests/minifier.js
```diff
@@ -639,6 +677,37 @@
     input = '<style>alert("foo     bar")    <\/style>';
     output = '<style>alert("foo     bar")<\/style>';
     equal(minify(input, { collapseWhitespace: true }), output);
+
+    // Collapse whitespace
+    var elements;
+    var element;
+    var i;
+
+    // Trim
```
kangax, line 686: Can we separate these into a separate test, for better readability and maintenance?

tests/minifier.js
```diff
@@ -639,6 +677,37 @@
...
+      input = '<' + element + '> foo </' + element + '>';
+      output = '<' + element +'>foo</' + element + '>';
+      equal(minify(input, { collapseWhitespace: true }), output);
+
+      input = '<' + element[0].toUpperCase() + element.slice(1) + '> foo </' + element + '>';
+      output = '<' + element +'>foo</' + element + '>';
+      equal(minify(input, { collapseWhitespace: true }), output);
+    }
+
+    // Collapse
```
kangax, line 699: Also this one

tests/minifier.js
```diff
@@ -769,6 +889,34 @@
     output = '<a href="#"></a><div>Well, look at me! I\'m a div!</div>';
 
     equal(minify(input, { html5: false }), output);
+
+    var elements;
+    var element;
+    var i;
+
+    // Unary
```
kangax, line 897: I would move this in a separate test as well

tests/minifier.js
```diff
@@ -769,6 +889,34 @@
...
+    elements = [ 'basefont', 'br', 'img', 'input' ];
+    for (i = elements.length - 1; i >= 0; i--) {
+      element = elements[i];
+      input = '<' + element[0].toUpperCase() + element.slice(1) + '><div></div></' + element + '>';
+      output = '<' + element + '><div></div>';
+      equal(minify(input, { html5: false }), output);
+    }
+
+    // Non-unary
```
kangax, line 906: And this

kangax:
Thanks for digging into this. It looks good to me and sounds reasonable. I would just want to make tests more granular (left few comments).

kangax:
@duncanbeevers could you resubmit and address my last comment? thanks!

duncanbeevers:
Yeah, I'm fine with closing this. I poked around at some of the edge cases and wrote some more granular tests, but it's a little snipe-hunty. I wasn't actually experiencing any problems and just noted some spots where the code could be made more robust.

I'll scratch out a few more cases if I get the itch.

PR #244, opened 2014-08-24
Remove removable attrs with unspecified values

duncanbeevers:


tests/minifier.js
```diff
@@ -260,6 +260,11 @@
 
     input = '<img src="" alt="">';
     equal(minify(input, { removeEmptyAttributes: true }), '<img src="" alt="">');
+
+    // preserve unrecognized attribute
+    // remove recognized attrs with unspecified values
+    input = '<div data-foo class id style title lang dir onfocus onblur onchange onclick ondblclick onmousedown onmouseup onmouseover onmousemove onmouseout onkeypress onkeydown onkeyup></div>';
+    equal(minify(input, { removeEmptyAttributes: true }), '<div data-foo></div>');
```
kangax, line 267: The side effect here is that style selectors like `div[id]`, `[lang]`, etc. will stop working. But that's true for any redundant attribute removal and is a rare case anyway.

kangax:
Thanks @duncanbeevers!

tests/minifier.js
duncanbeevers, line 267: Right. This just make the functionality the same as if the parser had encountered `<div class=""></div>`, and also fixes a crashing error when the parser encountered a valueless `style` attribute such as `<div style></div>`.

Issue #129, opened 2014-02-25
Only remove blank lines...


ghost:
...rather than put all the tags one after another.

kangax:
Can you give an example?

ghost:
In addition to this very minified mode:

```
<!DOCTYPE html><html lang="en" class="no-js"><head><meta charset="utf-8"><meta http-equiv="X-UA-Compatible" content="IE=edge"><title>Carbon</title><meta name="title" content="Carbon"><meta name="description" content="A front-end framework."><meta name="apple-mobile-web-app-capable" content="yes"><meta name="apple-mobile-web-app-status-bar-style" content="black-translucent"><meta name="viewport" content="width=device-width, initial-scale=1"><link href="stylesheets/application.css" rel="stylesheet"><script defer src="scripts/application.js"></script><link href="images/icn-32x32.png" rel="shortcut icon"><link href="images/icn-152x152.png" rel="apple-touch-icon"></head><body></body></html>
```

This mode. Without empty lines:

```
<!DOCTYPE html>
<html lang="en" class="no-js">
  <head>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <title>Carbon</title>
    <meta name="title" content="Carbon">
    <meta name="description" content="A front-end framework.">
    <meta name="apple-mobile-web-app-capable" content="yes">
    <meta name="apple-mobile-web-app-status-bar-style" content="black-translucent">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <link href="stylesheets/application.css" rel="stylesheet">
    <script defer src="scripts/application.js"></script>
    <link href="images/icn-32x32.png" rel="shortcut icon">
    <link href="images/icn-152x152.png" rel="apple-touch-icon">
  </head>
  <body></body>
</html>
```

kangax:
So spaces are preserved, but newlines are stripped?

Because we already have "collapseWhitespace" option, which when disabled, does this:

``` html
<p>foo</p> <p>bar</p>

<p>baz</p>
```

stays "untouched":

``` html
<p>foo</p> <p>bar</p>

<p>baz</p>
```

ghost:
Exactly! Something like `removeEmptyLines: true`.

So this:

```
<p>foo</p> <p>bar</p>

<p>baz</p>
```

Would spit this:

```
<p>foo</p> <p>bar</p>
<p>baz</p>
```

LaurentGoderre:
[...]

PR #246, opened 2014-08-25
Added option to preserve a single line break on collapse

LaurentGoderre:
This improves legibility of the HTML without affecting the compression


kangax:
Can you please add a test case

LaurentGoderre:
Sure!

LaurentGoderre:
There you go!

duncanbeevers:
@LaurentGoderre Your needs mentioned in [wet-boew#5818](https://github.com/wet-boew/wet-boew/issues/5818) might be served by simply using the `maxLineLength` option recently-added to html-minifier.

As it stands, this PR slightly increases the size of the generated output after gzipping by increasing the total entropy of the input corpus.

You can see the difference in compressed output size by running

```
node benchmark.js
```

Before applying this PR, the html-minifier corpus is agressively gzipped down to 8821 bytes while afterwards it compresses to 8828 bytes.

This is admittedly a pretty negligible difference, but if the `maxLineLength` functionality already incorporated in the project meets your needs, I'd like to avoid taking any hit to the baseline compression performance.

LaurentGoderre:
I admit that I did not think about that. The maxLineLength will not help us but I,m wondering if we are not pursing something unrealistic

LaurentGoderre:
@duncanbeevers @kangax would you be opened to adding this as an option?

duncanbeevers:
I am definitely _not_ against adding it as an option.

Can you explain what problem this pull request alleviates?

LaurentGoderre:
It's a request we have from our <del>user</del> implementers. They want the compression but want to keep some form of readability. We currently use an html minifier that does this but it's Java based and it's soooo slow. This especially helps when dealing with autogenerated output (ex: Jekyll)

kangax:
So it's similar to https://github.com/kangax/html-minifier/issues/129 then?

LaurentGoderre:
Quite similar but this also gets rid of whitespace before and after the line breaks. I'm wondering if this would meet their needs

kangax:
@LaurentGoderre I'm happy to merge this in, but couple things first:
1. Let's rename `preserveLineBreak` to `preserveLineBreaks`. I think it sounds more natural and is consistent with our other options (`removeRedundantAttributes`, `removeAttributeQuotes`, etc.)
2. Let's add a bit more complex test (more markup, e.g. the one from #129), just to be sure everything works as expected.
3. I'm wondering how this affects performance, but considering this is opt-in, I suppose we don't have to worry about it for now (except that `.replace(lineBreakStamp, '\n');` happens all the time now, right? not just when option is `true`?)

LaurentGoderre:
Regarding #3, would you rather it be:

``` js
if (options.preserveLineBreaks) {
  str = str
    .replace(lineBreakBefore, lineBreakStamp)
    .replace(lineBreakAfter, lineBreakStamp);
    .replace(/[\t\n\r]+/g, ' ').replace(/[ ]+/g, ' ')
    .replace(lineBreakStamp, '\n');
} else {
      // strip non space whitespace then compress spaces to one
      return str.replace(/[\t\n\r]+/g, ' ').replace(/[ ]+/g, ' ')
}
```

LaurentGoderre:
Ok, #1 and #2 are done. How do you want to deal with #3? (Note that the logic changed a bit)

kangax:
I think it's good now. Thanks again!

LaurentGoderre:
Can you let me know when you publish to npm?

kangax:
Published!

LaurentGoderre:
Btw, what do we do about the `preserveLineBreak` vs `preserveLineBreaks`?

kangax:
Wait, did we fix it or not?

LaurentGoderre:
I thought I did but I probably screwed up my rebase....sorry about that

niftylettuce:
Line breaks don't seem to be preserved either way - they don't get stripped if this is `false` or `true`.  Not sure what to do here, my code still looks like this output after passing through `html-minifier`:

``` html
<p>Foo</p>


A whole bunch of unwanted 

Line

Breaks
</p>
```

PR #247, opened 2014-08-28
Do not treat draggable as a boolean attribute.

shoha:
While the values of draggable are "true" and "false" (and "auto"), it does not behave like other Boolean attributes and is actually classified as an [enumerated attribute](http://drafts.htmlwg.org/html/master/editing.html#the-draggable-attribute).

For example, currently this code

``` html
<div draggable="true"></div>
```

...would be minified as

``` html
<div draggable></div>
```

...which is actually equivalent to

``` html
<div draggable="false"></div>
```

This is obviously incorrect.


kangax:
Thanks @shoja. Do you mind resubmitting? Can't automatically merge.

shoha:
Sure, done.

kangax:
thanks!

Issue #219, opened 2014-07-31
Don't remove empty <script>'s with non-empty src (removeEmptyElements)


eush77:
Empty `<script>`'s with non-empty `src` attribute should probably be preserved, just like `<textarea>`'s.

kangax:
Hm, good point. Same for `<link>`'s.

eush77:
Actually I can't see the problem neither with `<link>`'s, nor with other void elements (like `<img>` or `<meta>`). It seems that HTMLParser takes care of that.

duncanbeevers:
I think this can be closed.

kangax:
thanks @duncanbeevers

PR #251, opened 2014-09-02
Empty script

duncanbeevers:
Empty script element with `src` attribute must be preserved.

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -306,8 +306,20 @@
     return false;
   }
 
-  function canRemoveElement(tag) {
-    return tag !== 'textarea';
+  function canRemoveElement(tag, attrs) {
+    if (tag === 'textarea') {
+      return false;
+    }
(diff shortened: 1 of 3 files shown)


duncanbeevers:
Fixes #219

src/htmlminifier.js
```diff
@@ -306,8 +306,20 @@
...
 
-  function canRemoveElement(tag) {
-    return tag !== 'textarea';
+  function canRemoveElement(tag, attrs) {
+    if (tag === 'textarea') {
+      return false;
+    }
+
+    if (tag === 'script') {
+      for (var i = attrs.length - 1; i >= 0; i--) {
```
kangax, line 315: Btw, I usually do `(var i = attrs.length - 1; i--; )` in cases like this

kangax:
@duncanbeevers Thanks! Can you please add a test as well.

PR #252, opened 2014-09-02
Short-circuit corpus scan

duncanbeevers:
Huge speed-ups by not scanning the entire html corpus when we only care about the beginning few characters.

Previously, the wikipedia benchmark was coming in around 7 seconds on my machine. With this change, it takes about 2 and a half seconds.

### DIFF
--- a/src/htmlparser.js
+++ b/src/htmlparser.js
@@ -165,7 +165,7 @@
       if ( !stack.last() || !special[ stack.last() ] ) {
 
         // Comment:
-        if ( html.indexOf('<!--') === 0 ) {
+        if ( /^<!--/.test( html ) ) {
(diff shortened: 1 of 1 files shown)


kangax:
That's crazy. I would never think regex match (even constrained by boundaries) is faster than indexOf.

duncanbeevers:
`indexOf` has to scan through the entire string in the case of no matches. The regular expression can terminate its search at the first mismatched character.

An alternative implementation might be to `substring` off the front of the corpus and just do a strict equality comparison to that, but the regexp implementation generates fewer (no) intermediate objects for garbage collection.

kangax:
Makes sense. Thanks!

PR #264, opened 2014-09-12
Added an option to prevent the escaping

LaurentGoderre:


kangax:
What happens with escaping? Does it hurt somehow?

LaurentGoderre:
If you are using JSON data in HTML `data-` attributes yes (as show in the test case)

kangax:
Yeah I can see that. I'm just not understanding how escaping could affect anything negatively, if it's normally OK to escape quotes anywhere in HTML source (including data attributes).

LaurentGoderre:
Maybe there's something else at play for my issue but in any case, converting

```
data='{"foo":"bar"}'
```

to

```
data="{&quot;foo&quot;:&quot;bar&quot;}`
```

makes it quite a big larger. The larger the JSON object, the bigger it gets

kangax:
I see what you mean but I do feel a bit iffy about adding support for this.

@duncanbeevers what do you think?

LaurentGoderre:
I found out what the underlying issue is. It incorrectly encodes the data

```
data='{"foo": " \"bar\" "}'
```

becomes

```
data="{&quot;foo&quot;: &quot \"bar\" &quot;}'
```

which is incorrect. It should double encode the \ and keep encoding the double quotes

LaurentGoderre:
The problem is this line

https://github.com/kangax/html-minifier/blob/gh-pages/src/htmlparser.js#L351

Is should probably be:

```
escaped: value && value.replace(/(^|.)"/g, '$1&quot;'),
```

@jeresig do you remember why the `\"` pattern is excluded?

jeresig:
Ugh, lord knowns, that was so long ago! What you have is probably more correct :)

duncanbeevers:
I agree we should not unnecessarily expand attribute values, and JSON is getting embedded everywhere these days. It's simply not that unusual.

Right now we support the `removeAttributeQuotes`, which can generate documents with heterogeneous attribute-quoting styles.

Rather than a big switch that turns on and off attribute value quoting, I feel we should attempt to generate output that best suits the input.

For example, an option like `optimizeAttributeValueQuotes` would encompass the current behavior of `removeAttributeQuotes`, as well as attempt to find an optimal representation for attribute values not eligible for an unquoted representation.

As it stands, I think this would be as simple as counting the number of double-quotes and single-quotes within an attribute value and quoting the attribute value using whichever there are fewer few of.

I'm not sure whether this means we first need to normalize attribute values to a canonical representation before determining their output form. Doing so would allow us to generate optimized representations of otherwise noisy attribute values.

``` html
<div data="{&quot;foo&quot;: &quot;bar&quot;}"></div>
<div data='{"foo": "bar"}'></div>
```

This seems like a reasonable direction to take attribute-quoting, and worthy of a spike, although I think the proposed implementation is a bit narrowly-focused and demands too much interaction from developers.

This is just my two cents, but I'd like the options we expose to match programmer intent, and not demand too much a-priori knowledge of the corpus to be compressed or the mechanics of how said compression will be accomplished. Instead, they should be able to flip a simple switch and get out lots of goodies, with a good level of confidence that we haven't buggered their whole document.

kangax:
@LaurentGoderre btw, this is the commit that introduced that replacement — https://github.com/kangax/html-minifier/commit/d8a9f5f85ae39de912ddffe86a7f28b2bdecea57

LaurentGoderre:
@kangax in my other PR (#268), I fix my immediate problem and we could take a look at this separately

XhmikosR:
@LaurentGoderre: this needs to be rebased.

LaurentGoderre:
Done!

LaurentGoderre:
It seems that by rebasing, my PR now has the dist of 7818ce6c1202eec20cc1eea42f6340393c065d99 it seems like it was not included. I wonder if that's what could have caused the test failing

LaurentGoderre:
Can we merge this until the escaping of attribute can be figured out?

PR #268, opened 2014-09-15
Fixed nested quotes not being properly escaped

LaurentGoderre:
This addresses the immediate problem from #264


LaurentGoderre:
Can we merge this to fix the immediate problem with JSON in attributes?

XhmikosR:
@LaurentGoderre: if you add tests which pass, I'd be glad to have a look and merge this.

LaurentGoderre:
There was already a test but adding a more robust one identified a problem with my logic so I updated both

XhmikosR:
@kangax: any objections in merging this?

kangax:
I am not 100% sure that this is the best course of action but it's better than what we currently have and I don't have time to look into it more so... Yeah, I'm ok with merging :)

XhmikosR:
OK, I guess we'll know if we break something in the process. :P

PS. @kangax Maybe you should release a v0.6.9? https://github.com/kangax/html-minifier/compare/v0.6.8...gh-pages?w=false

XhmikosR:
@LaurentGoderre: this seems to break some browser tests with FF 33 at least  http://kangax.github.io/html-minifier/tests/

LaurentGoderre:
How do I reproduce this locally? Also, this did not change the logic that is failing

LaurentGoderre:
@XhmikosR can we try merging #276 first. They could be a problem if the source and test do not match the dist folder

XhmikosR:
@LaurentGoderre: locally, you need Ruby and Jekyll as this is what GitHub pages use. But anyway I solved this by running `grunt`.

Issue #83, opened 2013-09-04
Support for custom syntax in ignored tags


mhulse:
Based on the code introduced in this PR:

[**New feature: Ignoring tags** &mdash; Specifically, ignoring strings that start with `<?` or `<%` and end with `?>` or `%>`.
](https://github.com/kangax/html-minifier/pull/82)

... @kangax suggests:

> Now we just need to figure out how to support custom syntax in ignored tags.

mhulse:
@kangax, to clarify this issue, does that mean that there should be the option for people to specify their own start/end tags for the ignored pattern match?

One situation I can think of might be opening/closing tags for something like Django templates:

``` html
{% block foo %}html  stuff goes here{% endblock foo %}
{# this is a single line server-side comment #}
{{ variable }}
```

... where the pattern to match/ignore would be based on `{%`, `{#` or `{{` (for opening tags) and `%}`, `#}` or `}}` (for closing tags)?

In other words, should the developer be allowed to pass in a regex/pattern to the code for ignoring tags of their choice? If so, should this custom regex/pattern override the existing "ignore tags" regex or should it get appended-to (i.e. on top of matching `<?`, `<%`, `?>`, and `%>`, it would also match these custom opening/closing tags: `{%`, `{#`, `{{`, `%}`, `#}` and `}}`)?

kangax:
> In other words, should the developer be allowed to pass in a regex/pattern to the code for ignoring tags of their choice?

Yes, I think that would make sense.

> If so, should this custom regex/pattern override the existing "ignore tags" regex or should it get appended-to

I think most definitely replaced. If I'm working with ruby, I don't care to do anything about "<?", "?>" strings. I would just want to say something like:

```
ignoreStartTag: /<\%+/,
ignoreEndTag: /\%>/
```

and only get those stripped.

The default ones we introduced are just a convenience for 2 most popular languages, and/or until we introduce custom syntax registering.

Does this sound reasonable?

mhulse:
> Does this sound reasonable?

Definitely! :smile:

I don't think it would be hard to introduce as a feature either; it should just be as simple as adding two new options and then using those options to search for start/end tags instead of the hard-coded regex variables.

Like you say, the defaults for the options would be regex pattern that searches for `<?`, `<%` and `?>`, `%>`).

If you're open to the help, I could probably tackle this feature this weekend. :octocat:

kangax:
[...]

PR #269, opened 2014-09-16
Support ignored tags as preprocessing pass

### LINKED ISSUE #261: Ignored tags around attributes gives an error
Files like:
&lt;% if (true) { %&gt;
&lt;div class="test"&gt;&lt;/div&gt;
&lt;%}%&gt;

works fine, and the contents of the server tags are ignored, but if I have the server tags around an attribute , they are not ignored:

&lt;div class="&lt;%=test||"en"%>"&gt;&lt;/div&gt;

this fails because of quotes inside quotes, but if would make more sense if the code was just ignored, just like when they are outside a tag.

duncanbeevers:
Fixes #261

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -534,6 +534,12 @@
         isIgnoring = false,
         t = new Date();
 
+    if (options.removeIgnored) {
+      value = value
+        .replace(/<\?[^\?]+\?>/g, '')
+        .replace(/<%[^%]+%>/g, '');
+    }
+
     function _canCollapseWhitespace(tag, attrs) {
       return canCollapseWhitespace(tag) || options.canCollapseWhitespace(tag, attrs);
     }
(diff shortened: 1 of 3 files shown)


duncanbeevers:
This patch also provided ~5% speedup on the wikipedia benchmark.

duncanbeevers:
The change to the existing test is due to ignored tags no longer being handled as tags per-se, but instead as opaque strings. After they're removed, the ordinary whitespace collapsing rules come into play, and those rules preclude collapsing out the central `whitespace` in this construct: `text``whitespace``text`

duncanbeevers:
This could be pretty easily adapted to deal with #83 as well.

src/htmlminifier.js
```diff
@@ -534,6 +534,12 @@
         isIgnoring = false,
         t = new Date();
 
+    if (options.removeIgnored) {
+      value = value.
```
kangax: can you please move `.` to the next line, for consistency

duncanbeevers:
Oh yeah, this also makes ignored-tag-removal a little more strict, demanding matching opening and closing tag pairs.

`<%` can only be closed by `%>` and `<?` can only be closed by `?>`.

kangax:
@duncanbeevers thanks!

PR #278, opened 2014-10-02
Add support for attributes containing '.'

sigvef:
This pull request adds support for attribute names that contain `.`. The changes in the `dist` folder are generated by `grunt`. Let me know if there are any issues!


kangax:
I suppose if we already allow ":", we might as well support "."

@duncanbeevers @XhmikosR if there's no objection from you guys, I'll pull this in.

XhmikosR:
Yeah, makes sense :+1:

duncanbeevers:
@sigvef I'm curious about your use-case. What dialect of HTML are you using that allows for `.` in tag names?

sigvef:
@duncanbeevers 

> What dialect of HTML are you using that allows for . in tag names?

Just plain old HTML5 (which allows `.` in tag _attribute_ names). The [syntax spec](http://www.w3.org/TR/html5/syntax.html#attributes-0) allows attribute names to consist of:

> one or more characters other than the space characters, U+0000 NULL, U+0022 QUOTATION MARK ("), U+0027 APOSTROPHE ('), ">" (U+003E), "/" (U+002F), and "=" (U+003D) characters, the control characters, and any characters that are not defined by Unicode.

Our specific use-case is to use `.` as a namespacing divider in attribute names, e.g. `<div company.module.parameter=123>..</div>`. Hope that makes sense!

duncanbeevers:
Thanks @sigvef. Much appreciated!

PR #280, opened 2014-10-03
Fixed issue with preserveLineBreaks and line breaks at start and end of tag

LaurentGoderre:


LaurentGoderre:
Is there any issue with this?

PR #291, opened 2014-11-01
Fix `.` in attribute name with appropriate regexp

duncanbeevers:
The original implementation used the overly-permissive `.` in its regular expression, where a literal period is needed.


duncanbeevers:
Thanks Juriy!

kangax:
Thank _you_! :)

PR #296, opened 2014-11-11
Enable lint and its output for cli

### LINKED ISSUE #197: --lint always fails
Try: `echo '<html></html>' | html-minifier --lint`

NikolaySl:
Fixes #197

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -711,7 +711,6 @@
       customAttrSurround: options.customAttrSurround
     });
 
-    lint && lint.populate();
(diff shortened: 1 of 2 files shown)


src/htmlminifier.js
```diff
@@ -711,6 +711,7 @@
       customAttrSurround: options.customAttrSurround
     });
 
+    lint && lint.populate();
```
kangax: Let's move this to cli.js as it's meant to be called separately

NikolaySl:
Moved

PR #299, opened 2014-11-13
Special-case svg tags, preserve case and slashes

### LINKED ISSUE #285: Change rules when inside an <svg> element
If I have inline `<svg>` elements in my page, I have to use `keepClosingSlash: true, caseSensitive: true` because otherwise HTMLMinifier will break the SVGs...

but I might not want to have those options enabled for my whole page.

it would be nice if HTMLMinifier could just force these options to `true` for anything inside an `<svg>` element.

duncanbeevers:
Fixes https://github.com/kangax/html-minifier/issues/285

This is a pretty heavy-handed approach in that we override the user-supplied `keepClosingSlash` and `caseSensitive` options when the minifier encounters an svg tag.

Options are stored in a stack, and creating the stack relies on `Object.create` which may not be supported in all our runtimes.

Imbalanced input can lead to an over-popped options stack and a very ungraceful failure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ How does HTMLMinifier compare to [another solution](http://www.willpeavy.com/min
 | `removeEmptyElements`          | [Remove all elements with empty contents](http://perfectionkills.com/experimenting-with-html-minifier/#remove_empty_elements) | `false` |
 | `lint`                         | [Toggle linting](http://perfectionkills.com/experimenting-with-html-minifier/#validate_input_through_html_lint) | `false` |
 | `keepClosingSlash`             | Keep the trailing slash on singleton elements                            | `false` |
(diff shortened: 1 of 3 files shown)


duncanbeevers:
Ping @kangax

luisrudge:
still broken.

kangax:
I like it! Couple things:
- I think we shouldn't rely on `Object.create`
- If we're "overwriting" `keepClosingSlash` and `caseSensitive` behavior, let's mention it in the docs, to avoid any misunderstanding (I guess in README since we don't have docs docs).

duncanbeevers:
Updated with manual object copy in lieu of `Object.create`.
Added documentation of behavior to README.

kangax:
Thanks @duncanbeevers!

duncanbeevers:
This may warrant a `MAJOR` change.

Issue #300, opened 2014-11-16
benchmark.js not working on Windows?


XhmikosR:
I'm getting this

```
C:\Users\xmr\Desktop\html-minifier>node -v && npm -v
v0.10.33
1.4.28

C:\Users\xmr\Desktop\html-minifier>node benchmark.js

Processing... html-minifier

C:\Users\xmr\Desktop\html-minifier\benchmark.js:80
            var minifiedSize = stats.size;
                                    ^
TypeError: Cannot read property 'size' of undefined
    at C:\Users\xmr\Desktop\html-minifier\benchmark.js:80:37
    at Object.oncomplete (fs.js:107:15)
```

Maybe it's due to gzip not available on Windows?
/CC @kangax

kangax:
No idea actually. /cc @duncanbeevers

XhmikosR:
Using https://github.com/kangax/html-minifier/compare/benchmark2, I'm getting:

```
C:\Users\xmr\Desktop\html-minifier>node -v && npm -v
v0.10.35
1.4.28

C:\Users\xmr\Desktop\html-minifier>node benchmark.js
Processing... html-minifier

C:\Users\xmr\Desktop\html-minifier\benchmark.js:89
              throw new Error('There was an error reading ' + minifiedFilePath
                    ^
Error: There was an error reading benchmarks\generated\html-minifier.min.html
    at C:\Users\xmr\Desktop\html-minifier\benchmark.js:89:21
    at Object.oncomplete (fs.js:108:15)
```

/CC @kangax @duncanbeevers: let me know if that works for you; if so we can merge it since it's the proper way to handle dirs cross-platform.

About this issue, I think it's due to `gzip`; Windows doesn't have it.

duncanbeevers:
I haven't tried this out as I don't have a Windows rig handy.

I'm all-for cross-platform path handling, though I haven't really had to deal with it much myself.

@XhmikosR 
I see you've changed some lines to look like this:

``` javascript
path.join('benchmarks/generated/', fileName) + '.html.gz';
```

Is it somehow safe to use `/` within those sequences, but not to simply concatenate the path strings?

XhmikosR:
I pushed that branch a better fix; now everything works on Windows too.

Please try it on a non Windows OS just to be safe.

@duncanbeevers: The plan was to add the whole string in `path.join`, but previously I had other issues so it didn't work. Now it should work fine.

```
Processing... html-minifier
[...]

PR #301, opened 2014-11-16
Throw when fs.stat callback provides error

duncanbeevers:
In relation to #300 

There seems to be some error getting file stats on the minified file, probably due to some problem further up the pipeline, resulting in a missing file or something.

This PR just increases the visibility of the error, though you may want to dig into `err` and see if it contains any meaningful information rather than just throwing it away and exploding like this PR does.

@XhmikosR Let me know if this sheds any light on the issues you're seeing on Windows.

### DIFF
--- a/benchmark.js
+++ b/benchmark.js
@@ -77,6 +77,10 @@ function test(fileName, done) {
 
           // Open and read the size of the minified output
           fs.stat(minifiedFilePath, function (err, stats) {
+            if (err) {
+              throw new Error('There was an error reading ' + minifiedFilePath);
+            }
+
             var minifiedSize = stats.size;
             var minifiedTime = new Date() - startTime;
 


XhmikosR:
```
C:\Users\xmr\Desktop\html-minifier>node benchmark.js

Processing... html-minifier

C:\Users\xmr\Desktop\html-minifier\benchmark.js:81
              throw new Error('There was an error reading ' + minifiedFilePath
                    ^
Error: There was an error reading benchmarks/generated/html-minifier.min.html
    at C:\Users\xmr\Desktop\html-minifier\benchmark.js:81:21
    at Object.oncomplete (fs.js:107:15)
```

FYI, we should probably take care of these too (found with ESLint)

```
   28:0   error  Missing "use strict" statement               strict
   29:9   error  Don't use octal: '\0'. Use '\u....' instead  no-octal-escape
   29:29  error  Don't use octal: '\0'. Use '\u....' instead  no-octal-escape
   29:43  error  toKb was used before it was defined          no-use-before-define
   32:0   error  Missing "use strict" statement               strict
   33:9   error  Don't use octal: '\0'. Use '\u....' instead  no-octal-escape
   33:29  error  Don't use octal: '\0'. Use '\u....' instead  no-octal-escape
   33:43  error  toKb was used before it was defined          no-use-before-define
   36:0   error  Missing "use strict" statement               strict
   40:0   error  Missing "use strict" statement               strict
   43:9   error  Don't use octal: '\0'. Use '\u....' instead  no-octal-escape
   43:50  error  Don't use octal: '\0'. Use '\u....' instead  no-octal-escape
   46:0   error  Missing "use strict" statement               strict
   47:9   error  Don't use octal: '\0'. Use '\u....' instead  no-octal-escape
   47:29  error  Don't use octal: '\0'. Use '\u....' instead  no-octal-escape
   50:0   error  Missing "use strict" statement               strict
   66:20  error  Expected error to be handled                 handle-callback-err
   71:26  error  Expected error to be handled                 handle-callback-err
   89:42  error  Expected error to be handled                 handle-callback-err
  111:1   error  Missing "use strict" statement               strict
```

PR #318, opened 2014-12-23
Forward user-supplied options to uglifyjs

### LINKED ISSUE #303: Minify JS option "output" appears to be clobbered
Hi,
When using a config file with MinifyJS option, the "output" member appears to get over written instead of merged eg 'minify.json' config file contains:

``` json
"minifyJS": {"mangle": true, "output": {"comments": true}},
```

And html-minifier is invoked as follows:

``` console
$ html-minifier --config-file minify.json some-file.html
```

As a workaround I have patched my copy of htmlminifier.js as follows:
Line 1009:

``` js
    //options.output = { inline_script: true };
    if (typeof options.output !== 'object') {
      options.output = { };
    }
    options.output.inline_script = true;
```

duncanbeevers:
Fixes #303

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -464,7 +464,9 @@
       options = { };
     }
     options.fromString = true;
-    options.output = { inline_script: true };
+    var outputOptions = options.output || {};
+    outputOptions.inline_script = true;
+    options.output = outputOptions;
 
     try {
       // try to get global reference first


XhmikosR:
The patch looks good. Not sure if it makes sense, but I'll leave that to you and @kangax to decide <3

kangax:
LGTM. Can we make a unit test for this as well?

PR #320, opened 2014-12-26
Fixed reference to missing package & added dependency on htmlparser.

AshKyd:
This package doesn't seem to run under Node because it's looking for htmlparser locally, which doesn't exist. This fixes that issue and adds a dependency in package.json.


XhmikosR:
1. The html parser used here is modified.
2. The PR is already based on old master.

For more details wait for @kangax

kangax:
Yes, this is not the parser we use. In fact, our parser is not a standalone npm package as far as I know. It's supposed to reference it locally.

AshKyd:
My apologies, I seem to have misunderstood what's going on here. The tests were passing with this change, but it looks like it might have been a fluke rather than desired behaviour.

I also can't reproduce the issue I was having on NPM, so I'm closing this one. Edit: looks like it's specifically an issue with Browserify not being able to find the file. I'll look into it further and I'll come back to you with another PR if I work out what's going on

PR #321, opened 2014-12-28
Add application/ecmacript script type

silverwind:
I think the official ecmascript mime-type is missing here. Ref: [MDN](https://developer.mozilla.org/en/docs/Web/HTML/Element/script#attr-type)


kangax:
Oh thanks, I missed it :)

Issue #297, opened 2014-11-11
customAttrCollapse not working


thany:
I'm trying to use `customAttrCollapse` to do exactly what it's supposed to do, as far as I can tell from the readme. To collapse whitespace from a custom attribute. This is how I'm using it:

``` js
options: {
  collapseWhitespace: true,
  collapseBooleanAttributes: false
  removeComments: true,
  removeAttributeQuotes: true,
  removeScriptTypeAttributes: true,
  removeStyleLinkTypeAttributes: true,
  removeOptionalTags: true,
  customAttrCollapse: /ng\-class/
}
```

Example input:

``` html
<div ng-class="{ disabled: !charger_subgroup[features._name],
                recommended: charger_subgroup[features._name].recommenDation &gt; 0,
                notrecommended: !data.randomcars.count && data.recommendedTypecode != charger_subgroup[features._name].typeCode,
                active: data.selectedcharger.id === charger_subgroup[features._name].id }">
```

Output will be identical.
Please note: htmlmin is working fine apart from this function. The reason that this is identical, is simply because all other optimizations are not applicable, plus this is but a snippet from the rest of the file, which is minified perfectly.

I would expect the output to look like this:

``` html
<div ng-class="{ disabled: !charger_subgroup[features._name],recommended: charger_subgroup[features._name].recommenDation &gt; 0,notrecommended: !data.randomcars.count && data.recommendedTypecode != charger_subgroup[features._name].typeCode,active: data.selectedcharger.id === charger_subgroup[features._name].id }">
```

Kinda bulky, I know, but I assumed this is exactly what `customAttrCollapse` is for.

duncanbeevers:
`customAttrCollapse`, as the documentation says, only strips newlines from attribute values.

I agree that your use-case makes sense. It looks like we could change the regular expression in use.
https://github.com/kangax/html-minifier/blob/f1754cd5607c75b1faa903a19f2b14d10b4ee9eb/src/htmlminifier.js#L253

from (remove newlines)

``` javascript
attrValue = attrValue.replace(/\n+/g, '');
```

to (remove newlines and instances of two-or-more whitespace characters)

``` javascript
attrValue = attrValue.replace(/\n+|\s{2,}/g, '');
```

duncanbeevers:
Testing this change with your input produces your expected output.

``` html
[...]

PR #331, opened 2015-01-23
fix for issue #297

theasmoth:
https://github.com/kangax/html-minifier/issues/297


duncanbeevers:
The [Travis CI failure](https://travis-ci.org/kangax/html-minifier/jobs/48011501) looks spurious, but this functionality still needs to be accompanied by a test case.

Would you kindly add a simple test case exercising this functionality to make sure we don't regress on it in the future?

PR #335, opened 2015-02-25
add better support to ignore ejs template code

tschwartz:
Hello @kangax,

I was hoping to PR the addition of `<%=` to your ignored elements but am not familiar with Travis. Is there something easy that I'm missing that caused this failure?

Tyler

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -74,7 +74,7 @@ module.exports = function(grunt) {
     uglify: {
       options: {
         banner: '<%= banner %>',
-        //compress: true,
+        compress: true,
         mangle: true,
         preserveComments: false,
         report: 'min'
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "grunt-contrib-concat": "0.5.x",
     "grunt-contrib-jshint": "0.11.x",
     "grunt-contrib-qunit": "0.5.x",
-    "grunt-contrib-uglify": "0.7.x",
+    "grunt-contrib-uglify": "0.8.x",
     "grunt-exec": "0.4.x",
     "grunt-jscs": "1.5.x",
     "load-grunt-tasks": "3.1.x",


tschwartz:
Hello @kangax,

I was hoping to PR the addition of `<%=` to your ignored tags but am not familiar with Travis. Is there something easy that I'm missing that caused this failure?

Tyler

tschwartz:
@kangax Can you confirm that the test was actually the issue?

XhmikosR:
@tschwartz: please squash your patches into one.

PS. No need to open and close the PR every time.

kangax:
Hm, was the failure caused by this change? it must have been, since our tests passed before. We also need a new test for added "<%=". Other than that, LGTM.

kangax:
On a second thought, I'm surprised we need "<%=" at all. Shouldn't `/<%/` catch it as well?

tschwartz:
My apologies for opening and closing the PR numerous times. I have this bad habit of clicking the that button instead of the 'comment' button. I'm not sure why the original regex doesn't allow it but as it stands I get failures on anything that is `<%=`, that said I can dig into that some more if you wish.

kangax:
That would be great! And in any case, we need an additional test that
includes "<%=", at least to see that it's failing.

tschwartz:
@kangax @XhmikosR After thinking about this issue I have a feeling that it's not as simple as I thought it was. There are instances of both `<%` and `<%=` that are ignored by the plugin however where it is not ignoring the `<%=` is when an attribute and it's dynamic value are wrapped in a conditional like so:

```
<a href="http://www.someurl.com"
  <% if (btn.title) { %>
    title="<%= btn.title %>"
  <% } %>
>Some text</a>
```

What I'll do is revert my changes, run the tests with no changes to verify that the original test is indeed not broken (it probably isn't), write a test for this case and then fix it.

tschwartz:
@kangax @XhmikosR I wrote the tests for this scenario however not only would the tests fail but the minification itself fails which is a problem. It would appear that the current functionality doesn't handle nesting of server-side tags within logic. I don't have the time to look into this now but would be happy to PR the test into a feature branch if you so desired and then look into it when I have more time.

kangax:
@tschwartz Defintely desired! If you can get to the bottom of it, that would be great. Thanks. I have very little time now (with work occupying most of it).

tschwartz:
@kangax Would you mind creating a project branch? I'm somewhat in the same boat with work so I don't have any timelines as to when this will be complete.

kangax:
What do you mean by "creating project branch"?

tschwartz:
I guess it would have been better to ask you if you had a branching strategy that was more complex than just PRing into gh-pages. If you were to create a branch off of gh-pages then I could PR into that which would make it easier for you to check it out or to collaborate with others on the same feature.

kangax:
Nope, we don't do anything like that. Just PR into gh-pages (anyone will
still be able to pull your fork and check it out; or even see on the web
through your `username.github.io/...`)

PR #341, opened 2015-03-15
Use concat-stream for interactive stdin support

eush77:
The minifier is not compatible with other CLI tools as it tries to read the entire stdin synchronously.

``` js
> html-inline -i index.html |html-minifier
ERROR: interactive stdin input not supportedevents.js:141
      throw er; // Unhandled 'error' event
            ^
Error: write EPIPE
    at Object.exports._errnoException (util.js:734:11)
    at exports._exceptionWithHostPort (util.js:757:20)
    at WriteWrap.afterWrite (net.js:753:14)
```

This PR fixes that by using streams and delegating the low-level logic of reading, decoding and concatenating to [`concat-stream`](https://github.com/maxogden/concat-stream).

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -28,6 +28,7 @@
 'use strict';
 
 var cli = require('cli');
+var concat = require('concat-stream');
 var changeCase = require('change-case');
 var path = require('path');
 var fs = require('fs');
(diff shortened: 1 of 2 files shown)


valeriangalliat:
:+1: for this, had the same issue on FreeBSD while piping stuff to html-minifier (didn't have the problem on Linux though).

BTW I "fixed" it temporarily with the following patch:

``` patch
diff --git a/cli.js b/cli.js
index 72c59e2..e5354d9 100755
--- a/cli.js
+++ b/cli.js
@@ -176,8 +176,7 @@ cli.main(function(args, options) {
           // stdin input - typically when invoking a script without any
           // input (for interactive stdin input).
           // If you were to just continue, you'd create a tight loop.
-          process.stderr.write('ERROR: interactive stdin input not supported');
-          cli.exit(2);
+          continue;
         }
         else if (e.code === 'EOF') {
           // Happens on Windows 7, but not OS X 10.8.3:
```

valeriangalliat:
@kangax Thanks for merging! Is it possible to release this on npm too? Thank you!

kangax:
Tagged, pushed. Cheers.

valeriangalliat:
Awesome! Still not up-to-date [on npm](https://www.npmjs.com/package/html-minifier) though (sorry for insisting). Or maybe the `npm publish` is done in some kind of automatic job, and it's just a matter of time?

kangax:
Ahhh, good point. Check now? Just did `npm publish`.

valeriangalliat:
All good! :smile:

PR #365, opened 2015-05-22
fix #364 cli.js does not create RegEx objects from regex strings

### LINKED ISSUE #364: cli.js does not convert regex strings into RegEx objects
I'm trying to use the CLI to minimize Polymer 0.9+ HTML that includes attribute bindings:

``` html
<my-element hidden$="{{hidden}}"></my-element>
```

However, the following arguments do not work:

``` bash
--custom-attr-assign "[\"\\\\$=\"]"
--custom-attr-assign "[\"/\\\\$=/\"]"
```

It appears the problem is that cli.js needs to explicitly construct RegExp objects from the argument strings. (Since JSON has no notion of regexes, JSON.parse() will not automatically convert strings of the form `"/searchString/"` into RegExp objects.)

I am submitting a PR with a fix.

hadfieldn:


XhmikosR:
@hadfieldn: please squash your patches.

hadfieldn:
@XhmikosR Please forgive my ineptitude with git. I'm not sure I'm squashing correctly. I'm going to branch a fresh repository and try again.

PR #367, opened 2015-05-27
fix #364 to create RegEx objects from regex strings in cli.js

### LINKED ISSUE #364: cli.js does not convert regex strings into RegEx objects

hadfieldn:


kangax:
Thanks @hadfieldn! I'd love to see couple unit tests for this but I don't think we have anything that's testing our CLI yet. If you could add, that would be great, I'm merging anyway.

hadfieldn:
Thanks @kangax. Not sure how soon I'll have time to add some cli tests, but I'll keep this on my to-do list.

Issue #294, opened 2014-11-09
Fails with browserify due 'require("./htmlparser")'


ibc:
A simple example:

```
$ npm install html-minifier
```
- test.js:

```
var minifier = require("html-minifier");

module.exports = minifier.minify;
```
- Run browserify:

```
$ browserify test.js
Error: Cannot find module './htmlparser' from '/tmp/test/node_modules/html-minifier/dist'
```

This is because the "main" entry in package.json of `html-minifier` points to `./dist/htmlminifier.js` (really? this is not the NPM way at all). And such a file includes:

```
  if (global.HTMLParser) {
    HTMLParser = global.HTMLParser;
  }
  else if (typeof require === 'function') {
    HTMLParser = require('./htmlparser').HTMLParser;
  }
```

Since the call to `require()` uses a relative path it is "captured" by browserify() which attemps to load a file ("htmlparser.js") that does not exist at all. Shouldn't it be `require('htmlparser')`?

jantimon:
Same here - any updates?

jantimon:
Same as #84

alexlamsl:
Fixed by https://github.com/kangax/html-minifier/commit/c4a6d6fd47af75200ef2e1e17c12c622387df0c5

PR #370, opened 2015-05-31
Fix browserify issues #84 #294

### LINKED ISSUE #84: Module "./htmlparser" not found error.
Looks like `global.HTMLParser` doesn't get set properly in some node.js environment. [This line](https://github.com/kangax/html-minifier/blob/a7e68a6f87c327fd0aed2d1261503539d64c9a46/dist/all.js#L393) fails and outputs this error message:

<blockquote>
Error: module "./htmlparser" not found from "./node_modules/rhtml/node_modules/html-minifier/dist/all.js"
</blockquote>

Please set `main` in `package.json` to actual source directory to better support NodeJS. You can distribute client-side library via bower.

jantimon:
This merge request allows using browserify. 

Fixes: #84 and #294

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,10 +59,11 @@
   },
   "files": [
     "dist",
+    "src",
     "cli.js",
     "sample-cli-config-file.conf",
     "CHANGELOG.md",
     "LICENSE"
   ],
-  "main": "./dist/htmlminifier.js"
+  "main": "src/htmlminifier.js"
 }


kangax:
How will this work without parser? https://github.com/kangax/html-minifier/blob/gh-pages/Gruntfile.js#L69

jantimon:
I don't understand your question - I didn't delete anything.

kangax:
I'm saying if we specify main to be src/minifier, how will it work if src/minifier can't work on its own (without parser)

jantimon:
The parser is required: `HTMLParser = require('./htmlparser').HTMLParser;` 

https://github.com/kangax/html-minifier/blob/gh-pages/src/htmlminifier.js

kangax:
Ahha! Makes sense then, merging.

jantimon:
Cool :)

PR #383, opened 2015-07-01
Allow JSON files and node modules as config files

### LINKED ISSUE #298: minifyJS option - wanting to pass object containing a function from config file to UglifyJS
Hi,
I use the clean multiline string hack and want to inform Uglify to keep comments starting like /*! 
(see http://eli.thegreenplace.net/2013/11/09/javascript-es-5-hack-for-clean-multi-line-strings)

It would seem that parsing of the minifier config file is too strict and is not allowing an object containing a function, hence the following:

something.conf has a line:
"minifyJS": {comments: function (node,comment) { return(comment.charAt(0) === '!'); } },

$ html-minifier --config-file something.conf mysource.html
Error: Cannot read the specified config file

Thanks

fregante:
Fixes #298

Untested because `npm i` is currently hanging for me…

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -195,17 +195,30 @@ cli.main(function(args, options) {
   }
 
   if (options['config-file']) {
+    var fileOptions;
+    var fileOptionsPath = path.resolve(options['config-file']);
     try {
-      var fileOptions = JSON.parse(fs.readFileSync(path.resolve(options['config-file']), 'utf8'));
-      if ((fileOptions !== null) && (typeof fileOptions === 'object')) {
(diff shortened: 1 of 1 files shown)


kangax:
@bfred-it did you have a chance to test it yet?

kangax:
@bfred-it ping

fregante:
I'll try next week, sorry for the delay

fregante:
Okay, I tested it and it works correctly. I also added more explicit error messages, I hope I didn't go overboard.

kangax:
Great, thanks!

PR #387, opened 2015-07-23
improve and simplify logs with wrapped methods in `cli` module

yuanchuan:
1. make the version info outputted in a separate line.
2. highlighting the "Error" word.

reference:  https://github.com/chriso/cli#plugins


kangax:
Oh nice, didn't know about these, thanks!

PR #407, opened 2015-09-20
Allow the choice of character for enclosing attrbutes

jonnochoo:
Added an option to allow the character used to enclose an attribute to be specified to be either double or single quote. By default it uses double quotes (so shouldn't break any existing functionality). If a character other a single quote is used, it defaults to use double quotes.

So to specify a single quote would look something like this...

```
minify('<div class=\'bar\'>foo</div>', { quoteCharacter: '\'' })
```

The reason for adding this is that in our code we minify our HTML, but in C# we have to then escape the doubles to do something like...

```
String html = "<div class=\"bar\">foo</div>;
```

So using single quotes gets us around having to do this (for the better part anyway).


kangax:
That's a pretty good idea! Thanks.

PR #410, opened 2015-09-25
Feature/custom attribute carriage return support

### LINKED ISSUE #297: customAttrCollapse not working

bryanerayner:
Add support for customAttribute newline collapsing which include carriage returns and white space.

Fixes #297


kangax:
Nice, thanks! 

I'm on a phone now, can't check, but I wonder if we handle regular attributes same way?

Also, what if we use trimWhitespace instead, unifying behavior across regular and custom attributes? trimWhitespace does \s replacement so maybe we should account for newlines in there instead?

bryanerayner:
Perhaps a better method would be trimWhitespace(), then perform the regex on the resulting string?

Issue #431, opened 2015-11-06
spellcheck attribute is considered as a boolean attribute


karthilxg:
`spellcheck="false"` gets converted as `spellcheck`

`spellcheck` [isn’t a boolean attribute](https://html.spec.whatwg.org/multipage/interaction.html#attr-spellcheck)

> https://github.com/kangax/html-minifier/issues/63#issuecomment-18635032
> 
> Variations of spellcheck="true" and spellcheck="" can be optimized into just spellcheck. Any other values for this attribute that are not equal to the literal string false map to the default state, meaning e.g. spellcheck="lolwat" or spellcheck=" true" (with leading space) can be optimized by just removing the attribute completely.

``` HTML
<!--Input-->
<input type="text" spellcheck="false" autocomplete="off" autocorrect="off" autocapitalize="off">

<!--Output-->
<input type="text" spellcheck autocomplete="off" autocorrect="off" autocapitalize="off">

```

karthilxg:
**Workaround:** Setting the `collapseBooleanAttribute` value to `false` fixes this issue temporarily.

``` js
 htmlmin: {
  collapseBooleanAttributes: false,
}
```

PR #434, opened 2015-11-12
Attribute spellcheck is not a boolean 

genintho:
Fixes kangax/html-minifier#431

if spellcheck is something different from "" it is removed.

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -137,6 +137,8 @@
         attrName === 'type' &&
         attrValue === 'text') ||
 
+        (attrName === 'spellcheck' && attrValue !== 'false') ||
+
         (tag === 'script' &&
         attrName === 'charset' &&
         !attributesInclude(attrs, 'src')) ||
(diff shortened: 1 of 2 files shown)


kangax:
Hold on a sec. Are you sure this is correct?

To quote the spec:

> The spellcheck attribute is an enumerated attribute whose keywords are the empty string, true and false. The empty string and the true keyword map to the true state. The false keyword maps to the false state. **In addition, there is a third state, the default state, which is the missing value default (and the invalid value default)**.
> 
> The true state indicates that the element is to have its spelling and grammar checked. **The default state indicates that the element is to act according to a default behaviour, possibly based on the parent element's own spellcheck state, as defined below**. The false state indicates that the element is not to be checked.

So with this change, we'd be changing spellcheck's true to default which doesn't seem identical.

genintho:
I have followed what was describe in the issue. It is bit confusing and yes it is different from what the spec is saying.

The case I can think of (which are not what my code is doing):
`spellcheck="false"` should stay untouched

`spellcheck="true"` should become `spellcheck=""` (empty string and true are triggering the same behavior)
`spellecheck=""` should stay untouched
`spellcheck` ???
`spellcheck="random"` should trigger a lint error

For my understanding, the following

``` html
<form spellcheck="false">
  <input type="text" spellcheck="false"/>
</form>
```

can safely be transform to

``` html
<form spellcheck="false">
  <input type="text"/>
</form>
```

But that would probably be hard to code :p

Issue #391, opened 2015-07-29
Remove ALL Whitespace Between Elements


trisys3:
How hard would it be to have an option to remove ALL whitespace between elements? Specifically, when we have whitespace between inline elements (and maybe inline-block elements), the browser preserves that whitespace. Most browsers seem to preserve a single space, no matter how many spaces there actually are. It is related to this [StackOverflow question](http://stackoverflow.com/questions/5078239/how-to-remove-the-space-between-inline-block-elements).

I believe this module preserves these links as well, as from testing this seems to be the case. This, however, leads to more space than necessary, especially if you take out the space anyway with hacks like `font-size: 0`. The whitespace-preservation seems to be coming from the `collapseWhitespaceSmart` function in the `src/htmlminifier.js` file, lines 43-82.

Would it be possible to either remove this whitespace as well, or have an option to remove it? I don't see many times when this would be desirable, and if it is you can always use `whitespace: pre`. The one reason I could see where this would be an issue is that the final HTML file renders differently in the browser. This would be more a user education issue, and I am not sure how to "fix" it.

How feasible would this be?

Issue #304, opened 2014-11-24
Remove all white spaces between any tags


artuska:
I have complete control over my HTML code and I really need to remove all spaces between all type of elements: block level and inline level — now white spaces collapses only between block level elements.

Is there a way to remove all spaces between all tags completely?

(I'm using [grunt-contrib-htmlmin](https://www.npmjs.org/package/grunt-contrib-htmlmin) v0.3.0 Grunt plugin which is a wrapper of HTML minifier)

I've found similar issues like #273, but I still got spaces between `<button>` or `<span>` tags.

kangax:
Can you give a small example for reproduction?

artuska:
Well, it is not a bug, there is nothing to reproduce, everything is working as expected.

I have `collapseWhitespace` option set to `true`.

If I have a bunch of any block level elements in my HTML: `div`, `p`, etc — they are collapsing leaving no spaces between them:

original:

```
<section>
    <header>Header</header>
    <div>Div</div>
    <p>Paragraph</p>
</section>
```

htmlmin:

```
<section><header>Header</header><div>Div</div><p>Paragraph</p></section>
```

If I have bunch of any inline level elements: `button`, `span`, `b`, `i`, etc. — they are collapsing with 1 space between them:

original:

```
<section>
    <button>Button</button>
    <em>Em</em>
    <span>Span</span>
</section>
```

htmlmin:

```
<section><button>Button</button> <em>Em</em> <span>Span</span></section>
```

So, I'm just asking for an option to collapse white spaces between all type of tags, not only block-level tags — option like `collapseWhitespace` but which is more aggressive:

original:

```
<section>
    <header>Header</header>
    <div>Div</div>
    <button>Button</button>
    <em>Em</em>
    <span>Span</span>
</section>
```

htmlmin:

```
<section><header>Header</header><div>Div</div><button>Button</button><em>Em</em><span>Span</span></section>
```

kangax:
No, I understand. I just wanted to have an example handy for testing.

kangax:
Marking as feature; we should definitely add this.

artuska:
Thank you, that would be great! Should I close this issue?

kangax:
Only after we fix it (I'll try to do it within the next month or so if no one beats me to it)

AleksueiR:
:+1:

Njke:
:+1: Any progress on that?

Njke:
[...]

PR #440, opened 2015-11-21
Add collapseInlineTagWhitespace option

### LINKED ISSUE #273: Add possibility to remove whitespace around certain inline tags
#43 made collapseWhitespace “smarter” about inline tags, so that e.g. the space in `<b>bold</b> <i>italic</i>` is preserved by the transformation.

However, one could want the “junk space” between two buttons or two images to be removed rather that preserved. He would get a cleaner DOM tree in the end and then set appropriate margins in CSS.

```
html-minifier --collapse-whitespace --some-other-introduced-option main.html
```

main.html:

``` html
<div>
    <button>Push me</button>
    <button>Push me</button>
</div>
```

Expected output:

``` html
<div><button>Push me</button><button>Push me</button></div>
```

Actual output:

``` html
<div><button>Push me</button> <button>Push me</button></div>
```

sapics:
This PR add option `collapseInlineTagWhitespace` which remove white-space around inline tag.
If you have some better option name or suggestions, I will fix :)
This PR would close #273, #304, #391.


kangax:
Thanks @sapics, I think this looks good :)

PR #452, opened 2015-12-27
remove undefined status variable

aryszka:


ifraixedes:
I've just got this issue too.

dead-claudia:
[It appears already fixed in gh-pages](https://github.com/kangax/html-minifier/issues/454#issuecomment-168791310), by simply moving it to where it was supposed to be.

PR #460, opened 2016-01-05
Fix #459

### LINKED ISSUE #459: Output not printing to stdout in v1.1.1
Title explains the problem.

I dropped a `console.log(e.stack)` and `console.log(output)` [here in this catch block](https://github.com/kangax/html-minifier/blob/gh-pages/cli.js#L185-L189), and found that [the output was being judged as a file, even when `undefined`](https://github.com/kangax/html-minifier/blob/gh-pages/cli.js#L178). Here's the stack I got from it:

```
TypeError: Path must be a string. Received undefined
    at assertPath (path.js:8:11)
    at Object.posix.resolve (path.js:426:5)
    at runMinify (/home/isiah/dev-projects/js/website/node_modules/html-minifier/cli.js:179:33)
    at ConcatStream.<anonymous> (/home/isiah/dev-projects/js/website/node_modules/concat-stream/index.js:36:43)
    at emitNone (events.js:72:20)
    at ConcatStream.emit (events.js:166:7)
    at finishMaybe (/home/isiah/dev-projects/js/website/node_modules/readable-stream/lib/_stream_writable.js:511:14)
    at endWritable (/home/isiah/dev-projects/js/website/node_modules/readable-stream/lib/_stream_writable.js:521:3)
    at ConcatStream.Writable.end (/home/isiah/dev-projects/js/website/node_modules/readable-stream/lib/_stream_writable.js:486:5)
```

dead-claudia:
Fix worked locally.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -175,7 +175,7 @@ cli.main(function(args, options) {
     if (minified !== null) {
       // Write the output
       try {
-        if (output !== null) {
+        if (output !== null && output !== undefined) {
           fs.writeFileSync(path.resolve(output), minified);
         }
         else {
--- a/cli.js
+++ b/cli.js
@@ -175,7 +175,7 @@ cli.main(function(args, options) {
     if (minified !== null) {
       // Write the output
       try {
-        if (output !== null && output !== undefined) {
+        if (output != null) {
           fs.writeFileSync(path.resolve(output), minified);
         }
         else {
--- a/.jshintrc
+++ b/.jshintrc
@@ -8,5 +8,6 @@
   "node":       true,
   "strict":     true,
   "undef":      true,
-  "unused":     true
+  "unused":     true,
+  "eqnull":     true
 }


cli.js
```diff
@@ -175,7 +175,7 @@ cli.main(function(args, options) {
     if (minified !== null) {
       // Write the output
       try {
-        if (output !== null) {
+        if (output !== null && output !== undefined) {
```
kangax: I like to just do `!= null` in cases like this

dead-claudia: Okay. I wasn't sure because of the [.jshintrc](https://github.com/kangax/html-minifier/blob/gh-pages/.jshintrc). I usually use `!= null` as well.

kangax: I think jshint ignores sloppy equal comparison when it's against `null` for this exact reason

dead-claudia:
@kangax I fixed the .jshintrc to permit that. You're welcome.

kangax:
Thanks :)

XhmikosR:
Why didn't you fix the check to check for `undefined` instead?

dead-claudia:
@XhmikosR They both represent nothingness, and this is a little less brittle.

XhmikosR:
Well obviously, but the strict comparison is preferred.

dead-claudia:
@XhmikosR Although this is just a matter of preference, I was just following @kangax's [instructions](https://github.com/kangax/html-minifier/pull/460#discussion-diff-48880404) (that's also why I added the linter setting).

XhmikosR:
Whatever.

PR #473, opened 2016-01-27
implement custom event attributes for minifyJs

alexlamsl:
Also added `removeTagWhitespace` to web and command-line interfaces as I overlooked them the last time (never used them before, so learning my way around the project as I go...)


alexlamsl:
Forgot to mention - took the liberty to refine the default regex to `/^on[a-z]{3,}$/`, based on the observation [here](https://developer.mozilla.org/en-US/docs/Web/Events) that none of the event names are below 3 characters.

kangax:
:+1:

Issue #435, opened 2015-11-17
minifyJS apply for normal content


skahack:
``` js
test('', function() {
  input = '<script></script>foo';
  output = '<script></script>foo';
  equal(minify(input, { minifyJS: true }), output);
});
```

```
Actual value:
<script></script>foo;
Expected value:
<script></script>foo
```

I think inherit a script tag to `foo`.

PR #474, opened 2016-01-28
unwind currentTag

alexlamsl:
Also addresses a similar case with `minifyCSS`


tests/minifier.js
```diff
@@ -987,6 +987,10 @@
   });
 
   test('script minification', function() {
+    input = '<script></script>foo';
```
kangax: What was happening with this before the fix?

alexlamsl: See #435 - it performs `minifyJS` on `foo` and gives `<script></script>foo;`

alexlamsl:
improved test case for `minifyJS` so it's more clear what is being tested

alexlamsl:
@kangax FYI the two tests were run against original `htmlminifier.js` and both failed

kangax:
Hm, I'm still not sure about the point of this. Why are script contents outside of script elements should be minified?

alexlamsl:
Without this PR, `minify('<style></style>this + is(not, code)', {minifyJS:true})` would alter those text, same with `minifyCSS`

So I am fixing it so that it does not do that anymore, hence the tests expecting the inputs to be untouched.

alexlamsl:
IOW, contents outside of script elements should _not_ be minified, but they are currently being minified on `master`.

kangax:
Oh wow, didn't know. Then it's a serious bug, of course, thanks!

Issue #442, opened 2015-11-24
Odd results when using ignoreCustomFragments and <!-- htmlmin:ignore -->


nejsimon:
Hello

I'm trying to minify a wordpress theme when deploying it to a server. I've added regex to ignore any `<?php ?>` blocks. Some tags open in one file and closes in another, such as the `<body>` tag in header.php and `</body>` in footer.php so I'm wrapping them in `<!-- htmlmin:ignore -->`. However this causes problems when a `<?php ?>` block is within the tag wrapped in `<!-- htmlmin:ignore -->`.

Input `<!-- htmlmin:ignore --><body <?php body_class(); ?>><!-- htmlmin:ignore -->`

Expected result `<body <?php body_class(); ?>>`.

Actual result `<body  htmlmin29190600104629993 >`

Testcase:

```
var minify = require('html-minifier').minify;
var result = minify('<!-- htmlmin:ignore --><body <?php body_class(); ?>><!-- htmlmin:ignore -->', {
    ignoreCustomFragments: [
        /<\?php((.|\r|\n)+?)\?>/gi,
    ],
});
console.log( result );
```

kangax:
Looks like a bug; in the meantime you can use `ignoreCustomFragments`

Issue #436, opened 2015-11-17
ignoreCustomFragments space problem


skahack:
``` js
test('my tests', function() {
  input = '<img class="{% foo %} {% bar %}">';
  output = '<img class="{% foo %} {% bar %}">';
  equal(minify(input, { ignoreCustomFragments: [ (/\{\%[^\%]*?\%\}/g) ] }), output);
});
```

```
Actual value:
<img class="{% foo %}{% bar %}">
Expected value:
<img class="{% foo %} {% bar %}">
```

men232:
+1

kmturley:
For a workaround you can use optional regex params. If you change:

```
ignoreCustomFragments: [ (/\{\%[^\%]*?\%\}/g) ]
```

to be:

```
ignoreCustomFragments: [ (/\{\%[^\%]*?\%\}(\s)?/g) ]
```

This will optionally allow a space after the first opening tag. So this kind of html tags will work until a fix is made in the library:

```
<section class="header{% if loggedin = 1 %} logged-in{% endif %}">
```

PR #477, opened 2016-01-31
fixes for htmlmin:ignore and ignoreCustomFragments

### LINKED ISSUE #464: ignoreCustomFragments over CLI
Is there a CLI-equivalent option for ignoreCustomFragments when calling html-minifier over command line?  Thanks for your help.

alexlamsl:
#436 is fixed by taking adjacent whitespaces of custom fragments into account
#442 is caused by order of operation for `htmlmin:ignore` and `ignoreCustomFragments`

Also fixed #464 while I'm at it :stuck_out_tongue_closed_eyes:


alexlamsl:
Oops - slipped in `new Date()` --> `Date.now()` as VS Code was bugging me about it.

It is a [well supported feature](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Date/now), but if you feel strongly about it I can back it out from the commit...

alexlamsl:
`remove-empty-elements` was checked when "safe" options are selected on the web interface, and that made life rather difficult when doing quick tests

dist/htmlminifier.js
```diff
@@ -1162,6 +1162,14 @@
     return text;
   }
 
+  function uniqueId(value) {
```
kangax, line 1165: What is this for?

dist/htmlminifier.js
```diff
@@ -1178,37 +1186,37 @@
         stackNoTrimWhitespace = [],
         stackNoCollapseWhitespace = [],
         lint = options.lint,
-        t = new Date(),
+        t = Date.now(),
```
kangax, line 1189: Yeah, that's cool. I'm ok with it.

dist/htmlminifier.js
alexlamsl, line 1165: 1. replaces `'htmlmin' + (Math.random() + '').slice(2)`, ensure that the identifier generated would not collide with input
   - the loop will almost never be executed more than once
2. replaces `<!-- htmlmin:temp -->` to avoid potential collision with input

alexlamsl, line 1165: By replacing `<!-- htmlmin:temp -->` with `<!--!deadb33f-->`, we also get to simplify `isIgnoredComment()` would may give a minor performance boost.

Also, this PR avoids any extra work done when `ignoreCustomFragments` or `<!-- htmlmin:temp -->` are not specified in the input, i.e. `uniqueId()` and `String.prototype.replace()` would never be called if none of this features are requested by the input.

kangax, line 1165: Hm, I thought it'd be pretty unlikely that random number would collide with user input. But I guess this doesn't hurt and is not in any frequently-executed code.

kangax:
@alexlamsl thanks for all these awesome fixes and improvements lately! much appreciated :)

Issue #346, opened 2015-04-02
remove comments and collapse whitespace (non-conservative)


WolfgangKluge:
If I remove comments and collapse whitespace, I get

``` html
halloworld
```

from

``` html
hallo <!-- ... --> world
```

I think this is not the correct behaviour. One of the two whitespace characters should be preserved

alexlamsl:
Hmm, regardless of `removeComments`, `collapseWhitespace` is misbehaving:

```
> minify('<div>hello <!-- cruel --> world</div>', {collapseWhitespace:true});
'<div>hello<!-- cruel -->world</div>'
```

Note the removed whitespaces. At the very least I would expect one of the two whitespaces to be kept.

Issue #215, opened 2014-07-30
Make comment removal happen before whitespace collapsing


rbro:
When running html-minifier, I noticed the following minor behavior.

Running

```
html-minifier --remove-comments --collapse-whitespace --conservative-collapse
```

the html:

```
<html>

<!--test-->

</html>
```

is translated into `<html>  </html>`.  Note, there are 2 spaces between the html tags.  Should there be 1 space instead?  It looks like the extra space is because of the comment that was removed.

Is it possible for the comments to be stripped first and then the whitespace collapsed?

Thanks for your help.

kangax:
Hm, when I run it through http://kangax.github.io/html-minifier/ I get 1 space only, but probably shouldn't be any at all.

rbro:
I was expecting 1 space because of the option --conservative-collapse.

If I remove the comment on my own first , then I get 1 space.  Since I'm running --remove-comments, I was expecting the same 1 space since comments should be removed and then whitespace collapsed.

I was able to get 2 spaces on http://kangax.github.io/html-minifier with these options:
- Remove comments
- Collapse whitespace
- Conservative whitespace collapse

stevenvachon:
Additionally, `--conservative-collapse` should remove whitespace within `<head>`.

c-vetter:
+1
I ran into this issue today as well.

@stevenvachon 
That suggestion is probably better placed in a new ticket.

@rbro
It might be a good idea to rename the ticket to something like "Make comment removal happen before whitespace collapsing" in order to keep it focussed and precise.

papandreou:
I just tried to upgrade from html-minifier 1.1.1 to 2.1.1, and the fix for this (8b19a64) seems to have caused a regression, according to `git bisect`.

It comes down to:

``` js
console.log(require('html-minifier').minify('<div><span></span> <!--# echo "foo" --> bar</div>', {
    collapseWhitespace: true,
    conservativeCollapse: true,
    ignoreCustomComments: [ /^#/ ]
}));
```

which used to output:

``` html
<div><span></span> <!--# echo "foo" --> bar</div>
```

but now comes out as:

``` html
<div><span></span><!--# echo "foo" --> bar</div>
```

It seems like `conservativeCollapse` is not being honored anymore in this particular case?

papandreou:
Another weird thing that changed with that commit: With the same options as above,

``` html
<body>
<!--[if !IE]>-->Not IE<!--<![endif]-->
</body>
```

now minifies to

[...]

PR #481, opened 2016-02-04
Misc fixes

alexlamsl:
Getting these out of the way before I have a go at #215 and #346 :sweat_smile:


alexlamsl:
IMHO, we can merge `benchmark.conf` and `sample-cli-config-file.conf` to save ourselves a bit of maintenance overhead...

dist/htmlminifier.js
```diff
@@ -848,7 +857,9 @@
       return attrValue;
     }
     else if (isMetaViewport(tag, attrs) && attrName === 'content') {
-      attrValue = attrValue.replace(/1\.0/g, '1').replace(/\s+/g, '');
+      attrValue = attrValue.replace(/\s+/g, '').replace(/[0-9]+\.[0-9]+/g, function(num) {
+        return (+num).toString();
```
kangax: Shouldn't `num` already be a string? This is redundant double-conversion, it seems.

dist/htmlminifier.js
```diff
@@ -829,7 +838,7 @@
     }
     else if (isUriTypeAttribute(attrName, tag)) {
       attrValue = trimWhitespace(attrValue);
-      if (options.minifyURLs) {
+      if (options.minifyURLs && !isCanonicalURL(tag, attrs)) {
```
kangax, line 849: :+1:

dist/htmlminifier.js
kangax: Oh, I see what you did there...

dist/htmlminifier.js
```diff
@@ -1351,6 +1362,9 @@
         }
         if (options.minifyJS && isExecutableScript(currentTag, currentAttrs)) {
           text = minifyJS(text, options.minifyJS);
+          if (text.charAt(text.length - 1) === ';') {
+            text = text.slice(0, -1);
```
kangax, line 1377: Nice, yet another byte saved :)

dist/htmlminifier.js
kangax: I would add a comment just to make it clear what's going on "0.9000" -> "0.9", "1.0" -> "1", "1.0001" -> "1.0001"

kangax:
> IMHO, we can merge benchmark.conf and sample-cli-config-file.conf to save ourselves a bit of maintenance overhead...

Sure, go for it.

alexlamsl:
variable renamed and comment added for meta-viewport optimisation

kangax:
\o/

kangax:
Couple nits for the future, `Object.create` not available in all environments but I guess it's been long enough and IE<9 is largely dead. I would also call `matchingMap` method `createMatchingMap` or `createMap`, but this is obviously a tiny nit.

dist/htmlminifier.js
duncanbeevers, line 1377: Perhaps `text.match(/;$/)`?

alexlamsl, line 1377: That returns a whole match object, whereas `/;$/.test(text)` returns a boolean.

Anyway, I can't see how the RegExp version would give any performance advantage here.

PR #484, opened 2016-02-05
fix issues with whitespaces and comments

alexlamsl:
Also addresses https://github.com/kangax/html-minifier/pull/481#issuecomment-180048365 and updated `devDependencies`


kangax:
Awesome work, as usual, @alexlamsl Thanks a bunch!

PR #487, opened 2016-02-07
fix minifyURLs command-line usage

alexlamsl:
Also updated `benchmark.js` to use `minifyURLs`, and post the latest numbers on `README.md`


alexlamsl:
benchmark

PR #488, opened 2016-02-08
removeOptionalTags and other small fixes

alexlamsl:
`removeOptionalTags` is redone according to [HTML5](http://www.w3.org/TR/html5/syntax.html#optional-tags)

There are two rules that are absent from this implementation:
- start tag `<colgroup>`, need to be preceded by `</colgroup>`
- start tag `<tbody>`, need to be preceded by `</tbody>`, `</thead>` or `</tfoot>`

We will always save an extra byte (`/`) removing the end tag than the start tag, so AFAICT these are unreachable cases.

Aside from the unit tests, I've also run `benchmark.js` and skimmed through the output for sanity.

Other small fixes are the ones I've spotted along the way.


alexlamsl:
benchmark

alexlamsl:
benchmark

alexlamsl:
Speaking from my own experience, I used to turn `conservativeCollapse` on to work around various cases where `collapseWhitespace` was too aggressive, so `removeOptionalTags` became a no-go as it adds weird whitespaces which messes up alignment etc.

With the previous fixes to `collapseWhitespace` and this combined, I have been testing with

```
collapseWhitespace: true
conservativeCollapse: false
removeOptionalTags: true
```

on one of my projects (on IE11) and apart from smaller files, nothing seems to be different so far :blush:

kangax:
It would be nice for us to track our progress on benchmark files through the commit history. That way we can see exactly how much we've come and how much we progress/regress. @alexlamsl what do you think? Doable?

alexlamsl:
So you want me to run the benchmarks after each commit in this PR?

Or do you mean an automated system like Travis CI? :sweat_smile:

kangax:
Yeah, I mean something automated that updates numbers in a table for each commit (or for each "significant" commit). We can then visualize that table and get a nice graph of our benchmark filesizes going down.

Something like this:

apple-pe

alexlamsl:
Interesting idea - let me think about it while I travel for the week :wink:

alexlamsl:
So if we let the `grunt` job updates the table on `README.md` or something like it does to the `dist` folder, may be we can then track the git history of that and draw us a nice graph?

alexlamsl:
Scrap that - I was thinking about how slow `benchmark.js` is, until I realised I'm not going to hit all the other competitors for every single commit (unless I wanna get banned :laughing:)

Yeah, I'll try and cook something up next week :smiling_imp:

src/htmlparser.js
```diff
@@ -52,7 +52,7 @@
   });
 
   // Empty Elements - HTML 4.01
```
kangax: Hm, we should probably update comments about "HTML 4.01" since we longer constrain ourselves to it.

src/htmlminifier.js
```diff
@@ -838,6 +962,16 @@
         if (currentTag === 'style' && options.minifyCSS) {
           text = minifyCSS(text, options.minifyCSS);
         }
+        if (options.removeOptionalTags && text) {
+          if (optionalStartTag === 'html' || optionalStartTag === 'body' && !/^\s/.test(text)) {
+            removeStartTag();
+          }
+          optionalStartTag = '';
+          if (optionalEndTag === 'html' || optionalEndTag === 'body' || (optionalEndTag === 'head' || optionalEndTag === 'colgroup') && !/^\s/.test(text)) {
```
kangax, line 948: Should we maybe put these in a map? Feels weird to have them listed like this.

src/htmlminifier.js
```diff
@@ -743,16 +866,27 @@
...
-
         var lowerTag = tag.toLowerCase();
         if (lowerTag === 'svg') {
           options = optionsStack.pop();
         }
 
+        tag = options.caseSensitive ? tag : lowerTag;
+
+        if (options.removeOptionalTags) {
+          if (optionalEndTag && optionalEndTag !== 'dt' && optionalEndTag !== 'thead' && (optionalEndTag !== 'p' || tag !== 'a')) {
```
kangax, line 843: Same here. What's so special about these particular tags?

kangax:
Perfect! Enjoy your travels ;)

src/htmlparser.js
alexlamsl: I was also tempted to remove the commented code with regards to block elements below...

alexlamsl:
The reason for those `t == 'a' | t == 'b'` not being turned into `createMap()` is:
- only two matches
- too many for imagination of variable names :sweat:

But I guess it's probably more readable that way, so I'll do that. I should probably add some comments at those locations to clearly which omission rules are they taking care of.

See you next week!

alexlamsl:
@kangax please take a look and see if this is more to your liking :smile:

src/htmlminifier.js
```diff
@@ -176,14 +173,7 @@
...
   // https://developer.mozilla.org/en/docs/Web/HTML/Element/script#attr-type
-  var executableScriptsMimetypes = createMap([
-    'text/javascript',
-    'text/ecmascript',
-    'text/jscript',
-    'application/javascript',
-    'application/x-javascript',
-    'application/ecmascript'
-  ]);
+  var executableScriptsMimetypes = createMap('text/javascript,text/ecmascript,text/jscript,application/javascript,application/x-javascript,application/ecmascript');
```
kangax: Hmmm... not very readable is my only concern. Everything else looks great. Maybe add support for passing either string or an array to createMap?

alexlamsl: I've pulled it out into `createMap()` and `createMapFromString()`

kangax:
\o/

alexlamsl:
And now I shall work on that history script... :wink:

kangax:
That would be amazing. And thanks for all the awesome work so far! :+1:

Issue #495, opened 2016-02-16
merge <%...%> and <?...?> features into ignoreCustomFragments


alexlamsl:
At the moment `<% %>` and `<? ?>` are handled by `htmlparser` as special cases.

But if we were to update the defaults of `ignoreCustomFragments`, we can remove this extraneous logic from the parser, with the added benefit of reducing associated corner cases in `handler.char()`, not to mention potential improvement in performance.

@kangax do you think it's worthwhile?

kangax:
Definitely a good idea!

Issue #489, opened 2016-02-08
ignoreCustomFragments: <%tag%>


nosovk:
Hello!
I use your template for building SendGrid templates.
There is <% text %> wrapper for their special tags (like <%asm_group_unsubscribe_url%> )
But somehow gulp build fails in that tag. I know that it's invalid, that's why I try to add ignore regexp to html parser.

[projects/gulpCommonTasks/tasks/inliner.js](https://github.com/dudeonthehorse/Email/blob/dev/gulpCommonTasks/tasks/inliner.js)
.pipe(htmlmin({collapseWhitespace: true, minifyCSS: true, ignoreCustomFragments:"/<\%[\sa-z_]+\%>/ig"}))

but htmlmin still try to minify that tag and fails.
Any ideas how to fix that?

alexlamsl:
`minify()` handles `<% %>` natively, so no need to use `ignoreCustomFragments`:

```
> minify('<div> blah <%asm_group_unsubscribe_url%> </div>', { collapseWhitespace: true });
'<div>blah<%asm_group_unsubscribe_url%></div>'
```

Works as expected for me - what particular HTML code are you trying which is failing for you?

alexlamsl:
With `master` rather than `1.1.1` on `npm` this gives:

```
> minify('<div> blah <%asm_group_unsubscribe_url%> </div>', { collapseWhitespace: true });
'<div>blah <%asm_group_unsubscribe_url%></div>'
```

(Note the extra whitespace after `blah`.) Is this the desirable behaviour you are looking for?

alexlamsl:
@nosovk can you provide a more concrete example for the failure?

alexlamsl:
@nosovk I believe the issue is resolved, but do re-open this issue if that is not the case.

PR #496, opened 2016-02-16
addresses #495 and then some

alexlamsl:
In the process of refactoring for #495, I've discovered a bug when `conservativeCollapse` and `preserveLineBreaks` are used on elements with only `\n` on one side but whitespaces on the other.


alexlamsl:
I think by integrating the default value into `ignoreCustomFragments` in the docs, we also reduce the number of instances where users didn't realise `<?php ?>` is already accounted for, like #489 and others before it...

alexlamsl:
Now that I've worked with `ignoreCustomFragments` for a while, it's starting to look like a golden hammer :laughing: 

On a more serious note, contemplating on whether to use this for IE Conditional Comments...

alexlamsl:
Checked with `node benchmark` and no noticable regression in sizes or performance has been detected.

alexlamsl:
Travis CI tripped :confused:

dist/htmlminifier.js
```diff
@@ -618,13 +593,10 @@
     return createMap(values.split(/,/));
   }
 
-  // array of non-empty element tags that will maintain a single space outside of them
-  var inlineTags = createMapFromString('a,abbr,acronym,b,bdi,bdo,big,button,cite,code,del,dfn,em,font,i,ins,kbd,mark,math,q,rt,rp,s,samp,small,span,strike,strong,sub,sup,svg,time,tt,u,var');
-
-  function collapseWhitespaceSmart(str, prevTag, nextTag, options) {
+  function collapseWhitespace(str, preserveLineBreaks, conservativeCollapse, trimLeft, trimRight, collapseAll) {
```
kangax: 6 args is kinda scary; can we make it prettier? :)

dist/htmlminifier.js
```diff
@@ -634,24 +606,32 @@
...
 
     return lineBreakBefore + str + lineBreakAfter;
   }
 
+  // array of non-empty element tags that will maintain a single space outside of them
+  var inlineTags = createMapFromString('a,abbr,acronym,b,bdi,bdo,big,button,cite,code,del,dfn,em,font,i,ins,kbd,mark,math,q,rt,rp,s,samp,small,span,strike,strong,sub,sup,svg,time,tt,u,var');
+  var selfClosingInlineTags = createMapFromString('comment,img,input');
+
+  function collapseWhitespaceSmart(str, prevTag, nextTag, options) {
+    var trimLeft = prevTag && !selfClosingInlineTags(prevTag) && (options.collapseInlineTagWhitespace || prevTag.charAt(0) !== '/' || !inlineTags(prevTag.substr(1)));
```
kangax: Just need a little better formatting here (for easier reading)

dist/htmlminifier.js
```diff
@@ -1508,8 +1531,8 @@
           }
           optionalStartTag = '';
           // </html> or </body> may be omitted if not followed by comment
-          // </head> or </colgroup> may be omitted if not followed by space or comment
-          if (optionalEndTag === 'html' || optionalEndTag === 'body' || (optionalEndTag === 'head' || optionalEndTag === 'colgroup') && !/^\s/.test(text)) {
+          // </head>, </colgroup> or </caption> may be omitted if not followed by space or comment
+          if (optionalEndTag === 'html' || optionalEndTag === 'body' || (optionalEndTag === 'head' || optionalEndTag === 'colgroup' || optionalEndTag === 'caption') && !/^\s/.test(text)) {
```
kangax: Is scoping with parens necessary here?

alexlamsl: If you mean the parens around `head`, `colgroup` and `caption`, then yes, due to precedence of `&&` over `||`, i.e. `!/^\s/.test(text)` applies to all of those inside the parens, but not `head` or `body`.

But I guess another `createMapFromString` is in store here... :stuck_out_tongue_winking_eye:

alexlamsl:
@kangax PTAL

dist/htmlminifier.js
```diff
@@ -1568,8 +1595,12 @@
...
 
     if (uidAttr) {
-      str = str.replace(new RegExp('\\s*' + uidAttr + '\\s*', 'g'), function() {
-        return ignoredCustomMarkupChunks.shift();
+      str = str.replace(new RegExp('(\\s*)' + uidAttr + '(\\s*)', 'g'), function(match, prefix, suffix) {
+        var chunk = ignoredCustomMarkupChunks.shift();
+        return options.collapseWhitespace ? collapseWhitespace(prefix + chunk + suffix, {
+          preserveLineBreaks: options.preserveLineBreaks,
+          conservativeCollapse: true
+        }, true, true) : chunk;
```
alexlamsl, line 1603: OT: I wonder how far modern JavaScript engines can optimise this to:

```
if (options.collapseWhitespace) {
  str = str.replace(new RegExp('\\s*' + uidAttr + '\\s*', 'g'), function() {
    return ignoredCustomMarkupChunks.shift();
  });
}
else {
  str = str.replace(new RegExp('(\\s*)' + uidAttr + '(\\s*)', 'g'), function(match, prefix, suffix) {
    var chunk = ignoredCustomMarkupChunks.shift();
    return collapseWhitespace(prefix + chunk + suffix, {
      preserveLineBreaks: options.preserveLineBreaks,
      conservativeCollapse: true
    }, true, true);
  });
}
```

kangax:
Thanks! Just to confirm, this has no backwards-incompatible changes, right?

alexlamsl:
@kangax not intentionally, no :sweat_smile: 

For instance, the old rules for `<rtc>` etc. are kept to ensure compatibility between W3C and WhatWG versions.

alexlamsl:
@kangax on a related note, do we have plans for publishing `1.2.0` to `npm`?

kangax:
@alexlamsl done!

alexlamsl:
@kangax thanks :+1:

Issue #503, opened 2016-02-23
Removal of closing </p> tag breaks HTML


remcohaszing:
I have the following snippet of HTML:

``` html
<md-list-item ui-sref=".app-config">
  <md-icon md-font-icon="mdi-settings"></md-icon>
  <p translate>Configure</p>
</md-list-item>
```

After minification using `html-minifier@1.2.0`, this changes into:

``` html
<md-list-item ui-sref=.app-config>
  <md-icon md-font-icon=mdi-settings></md-icon>
  <p translate>Configure
</md-list-item>
```

This behaves seamingly random in the browsers I've tested (Chrome and Firefox). Most definitely the minified version doesn't behave as its unminified counterpart.

When using `html-minifier@1.1.1` everything still worked just fine.

I did a quick research and I think 0574066 is the culprit. Apparently it isn't safe to remove the closing tag `</p>` here.

alexlamsl:
AFAICT, `<md-list-item>` isn't part of HTML?

Anyway, you can switch off `removeOptionalTags` to avoid this issue when minifying general XML documents.

remcohaszing:
It's an AngularJS directive (https://material.angularjs.org/latest/api/directive/mdListItem).

I think custom elements are part of HTML. Otherwise several popular frameworks, such as AngularJS wouldn't be HTML compatible.

I do wish to remove optional tags, such as `<head>` and `<body>`.

alexlamsl:
From https://html.spec.whatwg.org/multipage/syntax.html#optional-tags

> A p element's end tag may be omitted if the p element is immediately followed by an address, article, aside, blockquote, details, div, dl, fieldset, figcaption, figure, footer, form, h1, h2, h3, h4, h5, h6, header, hgroup, hr, main, menu, nav, ol, p, pre, section, table, or ul element, or if there is no more content in the parent element and the parent element is an HTML element that is not an a, audio, del, ins, map, noscript, or video element.

So it certainly allows for removal of `</p>` if it's immediately followed by `</md-list-item>`. I have just tested by inspecting the DOM tree under IE11 and it adds `</p>` back as expected - what are those browsers you have mentioned behaved in this case?

alexlamsl:
(I assumed you've also applied at least `collapseWhitespace:true` since `</p>` won't be removed otherwise)

remcohaszing:
[...]

PR #508, opened 2016-02-25
fix crash in removeOptionalTags

alexlamsl:
also implement the optional start tag rules for empty element cases


alexlamsl:
@XhmikosR @kangax fixes for #503 is going to conflict with this, so if you can review and merge this first before I open another PR that would be great.

XhmikosR:
I'd still squash those 2 last patches, but otherwise it's much better than the previous PR :)

alexlamsl:
@XhmikosR squashed :pray:

alexlamsl:
rebased (`dist/htmlminifier.min.js` is always in the way...)

PR #509, opened 2016-02-25
fix whitespace trimming of ignoreCustomFragments

alexlamsl:


alexlamsl:
After this and #508 and a new PR for #503, all that is left is just a bunch of tests which involves all of them, which I guess I can just merge to `master` directly...

PR #514, opened 2016-02-29
make removeOptionalTags more conservative

alexlamsl:
As discovered in #503, we should treat all HTML tags around non-standard tags as mandatory.

They might get transformed by JavaScript libraries into various locations of the final DOM tree, which may or may not break the optional tag rules.


alexlamsl:
rebased

alexlamsl:
rebased again :ghost:

alexlamsl:
@kangax if you would publish `1.2.1` after this, I'd be grateful (to get beaten up by a new round of bug reports... :sweat_smile:)

tests/minifier.js
```diff
@@ -840,6 +840,54 @@
...
+    equal(minify(input, { collapseWhitespace: true, removeOptionalTags: true }), output);
+
+    input = '<html><body></body> </html>';
+    output = ' ';
+    equal(minify(input, { removeOptionalTags: true }), output);
+    output = '';
+    equal(minify(input, { collapseWhitespace: true, removeOptionalTags: true }), output);
+
+    input = '<html><body></body></html> ';
+    output = ' ';
```
kangax, line 886: Hm, it's probably more theoretical than anything, but I wonder if we really care about leaving space in cases like this...

alexlamsl, line 886: These tests are for the benefit of keeping track of this tricky logic than of any real world importance, I agree.

Pretty sure most people can't tell an empty screen from one with a whitespace on it :ghost:

PR #515, opened 2016-02-29
allow out-of-order placement in htmlmin:ignore & ignoreCustomFragments

alexlamsl:
Since the token we've chosen is unique, and they appear either as standalone comment or whitespace-aware text in `htmlmin:ignore` and `ignoreCustomFragments` respectively, we can safely append the index of the chunk to be read back and refer to afterwards.


alexlamsl:
rebased

PR #517, opened 2016-03-01
incorporate tests from PHPTAL

alexlamsl:
Ignored tests that:
- uses `tal:*` attributes
- evaluates PHP code
- uses `xmlns`
- uses `xml:space="preserve"`

Added comments to the ones which `html-minifier` gives different (better) output, and commented out the failed ones stating missing features.

### DIFF
--- a/tests/minifier.js
+++ b/tests/minifier.js
@@ -1818,4 +1818,105 @@
     equal(minify(input, options), output);
   });
 
+  test('tests from PHPTAL', function() {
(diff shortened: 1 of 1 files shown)


kangax:
I see you're having a lot of fun with this :)

alexlamsl:
Just a little :ghost: - it is always nice to learn about the dark corners of the technologies we use every day.

PR #518, opened 2016-03-01
fix CDATA comment removal

alexlamsl:
This started off as an adventure to tune performance, and ended up hunting down some bugs with CDATA optimisations :sweat: 

The performance improvement (~5% faster with `es6.html`) is all contained in `htmlparser.js` so hopefully won't be too confusing to review.

Unfortunately the boolean attribute optimisation is mixed in with the CDATA fixes in `htmlminifier.js`.

So if all of this turns out to be too confusing, give me a shout and I'll try and separate them into 3 PRs...


alexlamsl:
rebased

alexlamsl:
splitting this into smaller :cookie:

PR #519, opened 2016-03-01
handle whitespace within and around inline tags

alexlamsl:
inspired by the tests in PHPTAL:

```
<div>a <a href=#> b </a> c</div>
```

should minify to:

```
<div>a <a href=#>b </a>c</div>
```

before this PR:

```
<div>a <a href=#>b</a> c</div>
```

note the whitespace placement before `c`.

One case we cannot handle optimally is:

```
<div> <span> a </span> </div>
```

with this PR this turns to:

```
<div><span>a </span></div>
```

even though we should be able to remove that final whitespace safely.


alexlamsl:
Did I say we have a sub-optimal case? :smirk: 

(As it turns out, we don't need to flush `buffer` to `results` at all, so that helps.)

dist/htmlminifier.js
```diff
@@ -1503,7 +1514,22 @@
                 });
               }
             }
+            if (prevTag && inlineTextTags(prevTag.charAt(0) === '/' ? prevTag.substr(1) : prevTag)) {
+              text = collapseWhitespace(text, options, /(?:^| )$/.test(currentChars));
+            }
             text = prevTag || nextTag ? collapseWhitespaceSmart(text, prevTag, nextTag, options) : trimWhitespace(text);
+            if (!text && / $/.test(currentChars) && prevTag && prevTag.charAt(0) === '/') {
```
kangax: Are you sure we don't want `\s$` instead of `$`?

dist/htmlminifier.js
```diff
@@ -1503,7 +1514,22 @@
...
+              for (var index = buffer.length - 2, endTag = prevTag.substr(1); index >= 0 && _canTrimWhitespace(endTag); index--) {
```
kangax: I think `substr` is non-standard (`slice` is better)

kangax:
Do we know how this affects minifier performance?

dist/htmlminifier.js
alexlamsl: I thought of using `slice`, but then looking up and saw we use `substr` everywhere else :sweat_smile: 

Do you want me to replace all the usages?

alexlamsl:
I've run `benchmark.js` and no noticeable performance difference is observed.

In terms of minified sizes, they stayed almost the same - `eloquentjavascript` got <1KB bigger due to lots of:

```
<a>P </a>
```

`es6` got <1KB smaller due to consecutive trailing spaces being optimised.

dist/htmlminifier.js
alexlamsl: I'm thinking of collapsed whitespaces always getting turned into ` `- but I guess no harm making them`\s`

alexlamsl:
@kangax comments addressed :wink:

kangax:
Does it mean that "eloquentjavascript" was rendering incorrectly before? With some spaces removed?

tests/minifier.js
```diff
@@ -658,7 +672,7 @@
     equal(minify(input, { collapseWhitespace: true }), output);
 
     input = '<p> foo    <span>  blah     <i>   22</i>    </span> bar <img src=""></p>';
-    output = '<p>foo <span>blah <i>22</i></span> bar <img src=""></p>';
+    output = '<p>foo <span>blah <i>22</i> </span>bar <img src=""></p>';
```
kangax, line 675: Hm, why is this space necessary here?

alexlamsl, line 675: It's removing the space after `</span>` rather than the one before.

If there isn't any space before `</span>` then this wouldn't have happened.

alexlamsl:
Yes - those spaces were significant, so removing them packed those links too close to normal text after them.

XhmikosR:
@alexlamsl: are you sure this change doesn't break things? I notice differences in a project of mine, namely https://github.com/mpc-hc/mpc-hc.org (run it with `grunt server`). I had to turn on `conservativeCollapse` to get the same output as with the older version. Look for the spaces around the font awesome icons.

XhmikosR:
I also notice spaces before the end closing bracket, like

``` html
<a href=/acknowledgments/ >
```

I don't remember this one happening with the old version.

EDIT: happens with 1.2.0 too.

alexlamsl:
`<a href="/acknowledgements/">` should transform to `<a href=/acknowledgements/ >`, as `<a href=/acknowledgements/>` would cause the final slash to be confused as a closing slash for XML.

And I don't think this PR is the one that changes this IIRC...

XhmikosR:
But it's not XML.

alexlamsl:
Found it: https://github.com/kangax/html-minifier/commit/f8db419595ac792a37445e9a43f9e69f6548453c

The old behaviour is to retain the quotes as it'd cause some browsers to consume that as `/>`. The new behaviour saves one byte by removing the quotes but add a trailing whitespace to avoid parsing errors.

alexlamsl:
As in, `<a href=/acknowledgements/>` will get confused as `<a href="/acknowledgements"/>` in some browsers, hence `<a href=/acknowledgements/ >` is the shortest representation that is also safe.

I'll have a go at your project now - is there a specific link you'd want me to visit after `grunt serve`?

XhmikosR:
It's non standard behavior if "some" browsers behave like that. And HTML is not XML so the specs are clear. Anyway, that issue is a separate one. I'm more concerned about the whitespace issue I mention https://github.com/kangax/html-minifier/pull/519#issuecomment-199222850.

XhmikosR:
Just any page, but it's clearer in the homepage in the news post icons.

Note that I have already reverted to the old version in master so in order to see what I mean you will need to set grunt-contrib-htmlmin to `^1.0.0` or `1.1.0`.

XhmikosR:
Note that I have already reverted to the old version in master so in order to see what I mean you will need to set grunt-contrib-htmlmin to `^1.0.0` or `1.1.0`.

alexlamsl:
Thanks, I'll investigate now - I guess I can roll back and forth between the versions to see the difference as well...

alexlamsl:
Feels like an idiot, but... :sweat: 

```
> grunt serve

Warning: Task "serve" not found. Use --force to continue.

Aborted due to warnings.


Execution Time (2016-03-21 12:01:01 UTC)
loading tasks  3ms  ████████████75%
Total 4ms
```

Edit: I guess it's `grunt server`...

alexlamsl:
@XhmikosR I know I'm probably being annoying now, but is there a way I can test without installing ruby and jekyll and possibly more which I don't have on my PC? :cold_sweat:

alexlamsl:
Okay, so I visit https://mpc-hc.org/ on Chrome 49:
- grab `document.body.innerHTML`
- run it on http://kangax.github.io/html-minifier/ with most (safer) options and without `conservativeCollapse`
- paste the results back into `document.body.innerHTML`

And I compare it with another tab of the original site. The only thing I've noticed so far are the disappearing whitespace between `fa` icons and the following text.

Are there other discrepancies I should be paying attention to as well?

XhmikosR:
It's not possible to build without Ruby, sorry.

But what you see is the issue I described, yes. There might be more, but the font awesome ones were the obvious.

alexlamsl:
The whitespace here is being removed `</span> Downloads` - I'll track it down and try it and report back :wink:

PR #520, opened 2016-03-02
improve parser performance

alexlamsl:
Measured ~5% speed improvement on `es6.html` when running `benchmark.js`


alexlamsl:
The diff on github makes it look really dramatic, but in summary what I did:
- moved `html.indexOf('<')` from text handling forward, as all the other blocks depends on `<` being the foremost character to be processed, thus saving the transversal through all the case mismatches when text blocks are encountered
- record and update last tag name instead of repeatedly calling `stack.last()`, also prevents modifying the array object prototype which may have performance penalties

kangax:
Cool, feel free to merge (once it's rebased). As long as all tests are passing and code doesn't look too cryptic, I'm ok :D

PR #522, opened 2016-03-02
simplify isBooleanAttribute

alexlamsl:
Reduce line count on `isBooleanAttribute`, also added a test to clarify its case-insensitivity.


dist/htmlminifier.js
```diff
@@ -758,23 +758,11 @@
...
-    }
-
-    var attrValueEnumeration = enumeratedAttributeValues[attrName.toLowerCase()];
-    if (!attrValueEnumeration) {
-      return false;
-    }
-    else {
-      return (-1 === attrValueEnumeration.indexOf(attrValue.toLowerCase()));
-    }
+    return isSimpleBoolean(attrName) || attrName === 'draggable' && !draggableBoolean(attrValue);
```
kangax: I think `isBooleanValue` (instead of `draggableBoolean`) would be clearer here.

alexlamsl:
@kangax variable renamed, commit amended and pushed.

PR #523, opened 2016-03-02
improve whitespace handling when processing CDATA

alexlamsl:
One of the notable cases `removeCommentsFromCDATA` get wrong is this:

```
<script><!--
alert(1)
alert(2) --></script>
```

Expected:

```
<script>
alert(1)
</script>
```

Actual:

```
<script>
alert(1)
alert(2)</script>
```

This fixes other corner cases and their corresponding ones in `removeCDATASectionsFromCDATA` as well.


alexlamsl:
rebased

kangax:
Wait, why is `<script>
alert(1)
</script>` expected?

alexlamsl:
I've tested it in a browser and unless I put in `//-->` or `\n-->`, `alert(2)` won't be executed. I also get zero JavaScript error so the syntax isn't wrong or anything...

alexlamsl:
FWIW, Chrome (on Android) doesn't work at all with `<!--` and `-->` alone - it needs `//` or `/* */`, which I guess means it doesn't support the syntax, period. :disappointed:

kangax:
So the behavior among Chrome, WebKit, Mozilla, and Edge are different? Could you check all of them please?

alexlamsl:
Looks like both `<script><!--` and `<![CDATA` are a relic of the past. Safari behaves the same way as Chrome in that it doesn't support bare `--></script>` (so `//` or `/* */` is necessary to prevent syntax errors).

I'm still looking for a disposable OS instance to install and test Firefox on.

alexlamsl:
Right so we have two sets of behaviour for:

```
<script><!--
alert(1)
alert(2) --></script>
```

| alert(1) | syntax error |
| --- | --- |
| IE, Opera | Chrome, Safari, Firefox |

So my PR would match IE and Opera exactly, and for the others it will now run script which was ignored due to syntax errors.

alexlamsl:
I must confess, the more I'm testing with this, the more convinced I am to just remove these options altogether... :see_no_evil:

alexlamsl:
I think this is probably a tricky fix for very little gain.

Shall we settle for "if it ain't broke, don't fix it" for now? :sweat_smile:

kangax:
Yeah, I'm inclined towards that too.

alexlamsl:
Discarded.

Issue #499, opened 2016-02-20
IE Conditional Comments


alexlamsl:
So far I have encountered these forms:

```
<!--[if cond]>content<![endif]-->
<![if cond]>content<![endif]>
<!--[if cond]>-->content<!--<![endif]-->
<!--[if cond]><!-->content<!--<![endif]-->
```

where `content` can contain incomplete HTML markups.

Looking at #157, #225 & #274 - I propose the following behaviour:
- consume `content` as part of the comment for all forms of conditional comments, not just the first
- attempt to minify `content` just as we do with `processScripts`, fallback if error

alexlamsl:
Thinking about it some more, [Downlevel-revealed conditional comments](https://en.wikipedia.org/wiki/Conditional_comment#Downlevel-revealed_conditional_comment) are supposed to form part of the document without causing any parsing errors.

So my proposal is reduced to:
- minify `content` just as we do with `processScripts`, fallback if error
- fix `htmlparser` to allow for `content` to contain comments

kangax:
I don't remember entire context of downlevel-revealed comments but I think what you're saying sounds reasonable.

PR #525, opened 2016-03-02
fix whitespace between comments followed by tags

alexlamsl:
Run into this whilst fixing #499 - I'll let the tests speak for itself :ghost:


src/htmlminifier.js
```diff
@@ -796,6 +796,21 @@
       buffer.length = Math.max(0, index);
     }
 
+    function squashTrailingWhitespace(nextTag) {
```
kangax, line 790: Could you please add a comment as to what "squash" means here (since it's kind of ambiguous).

alexlamsl:
@kangax added comment and rebased - I hope it's not too confusing for you, and I welcome any suggestion to make the text clearer...

dist/htmlminifier.js
```diff
@@ -1322,6 +1322,24 @@
       buffer.length = Math.max(0, index);
     }
 
+    // look for trailing whitespaces from previously processed text
+    // which may not be trimmed due to a following comment or an empty
+    // element which has now been removed
+    function squashTrailingWhitespace(nextTag) {
+      var charsIndex;
+      if (buffer.length > 1 && /^(?:<!|$)/.test(buffer[buffer.length - 1]) && /\s$/.test(buffer[buffer.length - 2])) {
```
kangax: LGTM but after shortening this line (let's try to keep length <80)

PR #537, opened 2016-03-05
run browser-based tests

alexlamsl:
`grunt test` will now run `tests/*.html` using `phantomjs` as well.

Travis should fail at the moment due to #531.


alexlamsl:
Rebased - needs #531 in order to pass the new web tests.

alexlamsl:
`grunt-contrib-qunit` has been updated, but since this PR is getting rid of it anyway so I guess we can leave it alone for now...

alexlamsl:
With #531 merged, this is now good to go :sunglasses:

Issue #536, opened 2016-03-05
removeEmptyElements removes iframes


KenSharp:
It would appear that when removeEmptyElements is set to true, "empty" iframes are removed.

```
<iframe src='bacon.html'></iframe>
```

I know that it technically appears empty, but it really isn't. The attributes maketh the iframe.

This isn't true of frames:

```
<frame src='bacon.html'></frame>
```

...becomes:

```
<frame src=bacon.html>
```

Making the iframe "non-empty" stops it from being stripped:

```
 <iframe src='bacon.html'>iframe therefore I am</iframe>
```

Poor iframes.

alexlamsl:
On the bright side, I bet we've removed a lot of ads :smiling_imp:

KenSharp:
I wondered why embedded YouTube videos kept going missing....

I'm new to Github, I didn't realise every little thing I do would be logged against this ticket. 😳 I need to RTFM.

alexlamsl:
No harm for that - and thanks for taking care of it :+1:

PR #538, opened 2016-03-06
don't remove meaningful empty elements

alexlamsl:
#536 mentions `<iframe>` - I've further identified `<audio>` and `<video>`.


src/htmlminifier.js
```diff
@@ -481,18 +481,19 @@
...
-      for (var i = attrs.length - 1; i >= 0; i--) {
-        if (attrs[i].name === 'src') {
-          return false;
+    switch (tag) {
+      case 'audio':
+      case 'iframe':
+      case 'textarea':
+      case 'video':
+        return false;
+      case 'script':
```
kangax, line 498: Shouldn't video, audio and iframe act the same way as script? Only be removed when they don't have "src" attr?

alexlamsl, line 498: I was being conservative - but yeah let's go with yours.

Can you think of other tags while we are at this?

alexlamsl:
@kangax commit amended

kangax:
Perhaps `<object>`?

alexlamsl:
Assuming `<object data="...">` then?

alexlamsl:
Right, added `<applet code="...">` while we are at this.

alexlamsl:
... I've just discovered `<spacer>` as well :sweat:

alexlamsl:
Okay, looks like only Netscape and very old Mozilla browers support `<spacer>`, so I guess it's not a big deal...?

kangax:
HTML is hard :D

alexlamsl:
I'm actually thinking about `<i class="fa fa-camera-retro">` and same with `<span>` - they are commonly used to put out these font-based icons and `removeEmptyElements` would kill them.

kangax:
Not sure it's worth adding support for spacer. It's deprecated, no?

alexlamsl:
Probably not - it's one of those "no harm, let's throw in the kitchen sink" moment :ghost: 

I'll take it back out then...

alexlamsl:
So what are your thoughts on sparing `<i>` and `<span>` when `class` attribute is defined?

kangax:
> So what are your thoughts on sparing <i> and <span> when class attribute is defined?

That's expected. We should make it very clear in the docs. Ideally, they need to be ignored via config or something.

alexlamsl:
I guess that link to your blog post has enough warning signs on it.

alexlamsl:
`<spacer>` commit removed :wink:

kangax:
:+1:

PR #540, opened 2016-03-07
Omit auto-generated tags

alexlamsl:
Implemented an option to suppress synthetic tags generated by `htmlparser` from showing up in the minified output.

While it is useful for the parser to generate these signals for minifier to work within the correct context, they do have the effects of breaking browser hacks and other intentional markup omissions.

As it is guarded behind `ignoreSyntheticTags` and is turned off by default, we should be backward-compatible.

Also take out `cleanAttributes` from tests since the option no longer exists.


tests/minifier.js
```diff
@@ -1962,6 +1962,22 @@
     equal(minify(input, options), output);
   });
 
+  test('ignore synthetic tags', function() {
+    equal(minify('<p id=""class=""title="">x', { ignoreSyntheticTags: true }), '<p id="" class="" title="">x');
```
kangax: How would the output be different with `ignoreSyntheticTags: false`?

tests/minifier.js
```diff
@@ -410,103 +410,103 @@
 
   test('cleaning class/style attributes', function() {
     input = '<p class=" foo bar  ">foo bar baz</p>';
-    equal(minify(input, { cleanAttributes: true }), '<p class="foo bar">foo bar baz</p>');
+    equal(minify(input), '<p class="foo bar">foo bar baz</p>');
```
kangax, line 421: Nice, thanks!

tests/minifier.js
alexlamsl: It'd add `</p>` - I'll add a further test for clarification.

kangax: Thanks. Yeah, it's not really intuitive to me what `syntheticTags` mean. Could we perhaps say `autoGenerated` or `autoClosing`? Or is "synthetic" the actual term from the spec?

alexlamsl: I made that term up, and I'm notoriously bad at naming things :sweat_smile: 

`autoGenerated` sounds generic enough for possible further expansion - though in this case the default would be `true`, i.e. it'd be the opposite sense to the current setting. Is that okay?

kangax: how about `includeAutoGeneratedTags` that's true by default?

alexlamsl:
@kangax flag (and commit message) renamed, tests added

PR #541, opened 2016-03-07
handle tags with phrasing content model

alexlamsl:
HTML5 have strict definitions for content model, so things like `<p>` cannot have `<div>` inside for instance.

Auto-close such tags on the boundary of non-phrasing content.


alexlamsl:
Browsing around [HTML 4.01](https://www.w3.org/TR/html4/struct/global.html#h-7.5.4) doesn't seem to have the same restrictions, so confining this logic to `html5` mode.

alexlamsl:
Did a grep on https://html.spec.whatwg.org/ and populated the list of phrasing-content-only tags.

jshcrowthe:
Great work with the spec compliance here! I am using `html-minifier` (via gulp) for a while now and it has been great. However, in an application using web components, this addition broke me. 

Web Components have the unique nature that they can be put in any of the available [content categories](https://html.spec.whatwg.org/multipage/dom.html#kinds-of-content). In order for this to work with those using web components, tags that are not, by default, part of the HTML spec (and therefore recognized as  phrasing, flow, or another type) should be ignored.

I added the following test to https://github.com/kangax/html-minifier/blob/gh-pages/tests/minifier.js#L1191 and it failed:

``` javascript
  test('phrasing content with Web Components', function() {
    input = '<span><phrasing-element></phrasing-element></span>';
    output = '<span><phrasing-element></phrasing-element></span>';
    equal(minify(input, { html5: true }), output);
  });
```

devinus:
How can I turn this off?

devinus:
This is closing the tag even when there's a tag later on closing it, which is against the original intent.

PR #542, opened 2016-03-07
improve attribute parsing for conformance and performance

alexlamsl:
While following suggestion from @XhmikosR to include [Bootstrap](http://getbootstrap.com/css/) into benchmarks, I've discovered the `htmlparser` will grind to a halt with:

```
<a b=some://really.really/really/really/really-really-really-really-loooooong-content>
```

So fixed that, alongside the attribute parsing which was out-of-sync with start-tag parsing logic. In the process attribute parsing is now made conformant to the following documents:
- https://www.w3.org/TR/html4/intro/sgmltut.html#h-3.2.2
- https://html.spec.whatwg.org/multipage/syntax.html#attributes-2

With one notable exception - HTML5 attribute name allows `<` (but not `>`), and I couldn't for the ~~life of me~~ moment see a ~~reason~~ performant solution, so I've disallowed it for now.


alexlamsl:
So the story goes, `grunt test` used to take ~2s here.

As the HTML parser is generalised to handle non-alphanumeric attribute names, this goes to ~17s.

With the latest optimisation, this is now reduced to ~1.5s, so we get the cake and can eat it too! :birthday:

alexlamsl:
Online test suite (apart from #531) and `node benchmark` give it a clean bill of health as well.

dist/htmlminifier.js
```diff
@@ -285,8 +237,30 @@
       parseEndTag();
     }
 
-    function parseStartTag( tag, tagName, rest, unary ) {
-      var unarySlash = false;
+    function parseStartTag(input) {
+      var start = input.match(startTagOpen);
+      if (start) {
+        var match = {
```
kangax, line 243: Usually better to move var's out of the blocks; alternatively (and ideally) we should just switch to ES6 and let/const

alexlamsl, line 243: Noted for `var` hoisting.

As for `let`/`const` - I don't think we are quite ready for ES6 yet if we are to support a range of web browsers...

kangax, line 243: I wonder what percentage of people use minifier in a browser. Seems like a very very rare usecase (<1%) unless I'm missing something?

XhmikosR, line 243: Can't we just transpile the browser version? I have no experience with ES6, but it should be doable judging by other projects.

alexlamsl, line 243: I'd have no idea on this one. I can only say I haven't used it in any of my projects yet.

XhmikosR, line 243: I mean, we use Babel in Bootstrap v4-dev.

kangax, line 243: One option is to use Babel. Another option is to use Node (which supports a subset of ES6) and compile (deprecated?) browser version of minifier via browserify, which would — I hope — take care of ES6.

XhmikosR, line 243: It's just feels that this change is pretty big since everything will need to be rewritten.

Maybe wait a little more?

alexlamsl, line 243: @kangax @XhmikosR you two make up your mind and write it in a new issue - I'll pick it up from there :wink:

PR #543, opened 2016-03-07
allow XML tag names

alexlamsl:
Taking the [XML specification](http://www.w3.org/TR/xml/#NT-Name) and restrict it to within ASCII range, which seems reasonable enough a compromise until somebody wants their native language or emojis as tag names :see_no_evil:


kangax:
Hm, do all browsers parse/tolerate it well?

alexlamsl:
I've come across quite a few projects with these weird tag names - but yes, may be they all hit a brick wall at some point when they discover some browser breakage.

I will test against different browsers and report back.

alexlamsl:
Both IE11 and Chrome (48 on Android) seems to be happy with this:

```
<html>
<body>
<a.b c="d"></a.b>
<script>
alert(document.getElementsByTagName('a.b')[0]);
</script>
</body>
</html>
```

in that they both give me a pop-up with `[object HTMLUnknownElement]`.

kangax:
@mathias do you know of any problems with this? Safe to add?

kangax:
Whoops, I meant @mathiasbynens

mathiasbynens:
This is an improvement but it’s still not fully accurate. The relevant production is `NCName`: https://www.w3.org/TR/1999/REC-xml-names-19990114/#NT-NCName E.g. `elemænt` is valid. There’s an npm package for that™: https://www.npmjs.com/package/ncname / https://github.com/sindresorhus/ncname

Custom element names ([validator](https://mothereff.in/custom-element-name)) have some further requirements (like `-` being required) but we probably shouldn’t bother with those at all.

alexlamsl:
Didn't know modules with a bunch of `RegExp`'s exist. :open_mouth: 

I'll give this (or [xml-char-classes](https://github.com/sindresorhus/xml-char-classes/blob/master/index.js)) a go and see if I can implement this `QName` properly without performance regression.

alexlamsl:
@mathiasbynens @kangax done - added browserified module for online version as well, with fallback when module isn't available.

PR #544, opened 2016-03-08
minify URLs within CSS

alexlamsl:
Turns out `rebase` in [clean-css](https://github.com/jakubpawlowicz/clean-css/issues/740) can't quite achieve what we want here, so write a thin wrapper around `minifyCSS` to post-process the output instead.


alexlamsl:
For the record, clean-css doesn't consider this as a feature they would like to carry on their side.

https://github.com/jakubpawlowicz/clean-css/issues/740

Issue #547, opened 2016-03-08
Changelog for 1.3.0


alexlamsl:
Rather than spamming gitter, I guess I'll just update it here as we wait...
### Bug fixes
- messages cli.js (#498)
- extraneous whitespace inserted by `ignoreCustomFragments` (#502, #505, #506, #509, #513, #545, #552)
- `removeOptionalTags` on empty element crashes (#500, #508)
- `HTMLParser` halts on malformed markups (#177, #332, #507, #510, #512, #542)
- `removeComments` would give incorrect result if comment matches `ignoreCustomFragments` (#511, #515)
- `removeOptionalTags` too aggressive around non-HTML tags (#503, #514)
- incorrect `collapseWhitespace` within inline tags (#519)
- `removeTagWhitespace` does not work on the web (#528)
- `cli` cannot handle nested directories (#532, #533)
- `removeEmptyElements` always remove `<iframe>`, `<audo>` & `<video>` (#536, #538, #539)
- incorrect auto-closing of tags which permits phrasing content only (#541)
- extraneous whitespace inserted in attributes around custom fragments (#551, #553)
- `minifyJS` does not work the same way on the web (#531)
### Improvements
- minification of contents within conditional comments (#524)
- `collapseWhitespace` around comments and `removeEmptyElements` (#525)
- clean up `<script type="...">` (#521)
- support parsing of non-conventional attributes (#289, #534, #542)
- general performance (#520, #522)
- don't inject tags generated by HTML parser (#540)
- support parsing of non-standard tag names (#248, #543)
- migrate from `jscs` & `jshint` to `eslint` (#554)
- `grunt test` to cover browser tests (#537)

kangax:
@alexlamsl this is suuuuper helpful! You're amazing, thanks :heart: I'll tag and push the release with this list once we finalize all the existing PR's

alexlamsl:
IMHO we are good for `1.3.0` without #544:
- `benchmark.js` gives identical results before & after
- `clean-css` doesn't seem too enthusiastic about it

kangax:
Perfect, I'll tag and release as soon as I can! Thanks for all your help again.

alexlamsl:
Yay!

kangax:
So what's on the menu for 1.4.0? :)

alexlamsl:
About that... :smiling_imp:

alexlamsl:
[...]

PR #552, opened 2016-03-14
Add `removeCustomFragmentsWrap`option

yshterev:
If custom fragments are inside `src` or `href` attribute, the space breaks paths. 

Example:

``` html
<img src="{$assets_dist}img/logo.png" alt="Logo">
```

becomes

``` html
<img src="{$assets_dist} img/logo.png" alt="Logo">
```


alexlamsl:
I believe this has already been fixed - please check [online version](http://kangax.github.io/html-minifier/)

alexlamsl:
Fun fact - your new test case passes _without_ any changes to `htmlminifier.js`...

yshterev:
Hm.. yes it works from the fork but it does not work when I install from npm. 
Is it possible the npm package is not updated with latest changes ?

I know the test is a little weird :) .. I just wanted to verify my case will work..a legacy smarty project

alexlamsl:
The fix is merged but not published yet - follow #547 for when we do :wink: 

Thanks for reporting and working on it nonetheless. I'll close this for now and add it to our changelog, but if you find anything that doesn't work in the online version, please don't hesitate to help out!

yshterev:
Sure :thumbsup: Thanks for the quick response.

PR #554, opened 2016-03-15
replace jscs & jshint with eslint

alexlamsl:
@XhmikosR as per request, this is the migration to `eslint`.

As I'm completely new to this module, `.eslintrc.yml` is generated using `eslint --init` and feeding it the source files of this project, plus a few minor edits.


XhmikosR:
Great work! Only issue is with the format you went with for eslintrc. I'm used to the json. Also, are you sure the recommended ones don't conflict with the custom rules you added?

Generally, I just specify what I need, but if we can make this cleaner I'm all for it as we keep things simple. Having to go through all of those rules is a p.i.t.a. :/

alexlamsl:
Ah okay. I run into trouble and take note from Node.js and went with yaml, and all of those are generated by `eslint --init`. I only editted the offending rules but didn't add any new ones.

Let me start again with JSON with just `extends` and `env` and see if that'd fare better...

alexlamsl:
@XhmikosR right, 3rd try at this... forced push and with very minimal `.eslintrc.json` this time.

.eslintrc.json
```diff
@@ -0,0 +1,17 @@
+{
+  "env": {
+    "browser": true,
```
XhmikosR: I'd suggest that we specify the environment where needed and not globally.

XhmikosR:
@alexlamsl: apart from my comment, I'd go with `.eslintrc` and merge if everything passes. We can tweak the config later if needed.

alexlamsl:
@XhmikosR according to [here](http://eslint.org/docs/user-guide/configuring#configuration-file-formats), they claim `.eslintrc` to be deprecated

But if you insist, your wish is my command :angel:

XhmikosR:
No worries, it's a minor thing. :+1:

alexlamsl:
Turns out what I meant was `shared-node-browser` - the common subset of both...

alexlamsl:
Hmm, I'm confused - the test passed on my machine but not on Travis...

alexlamsl:
So I'll chalk that up as `node-chakracore`-specific behaviour...

kangax:
:+1:

Issue #557, opened 2016-03-16
Web Components are removed from phrasing-content-only blocks


jshcrowthe:
(Copy and paste from: https://github.com/kangax/html-minifier/pull/541#issuecomment-197554845)

Great work with the spec compliance here! I am using html-minifier (via gulp) for a while now and it has been great. However, in an application using web components, this addition broke me.

Web Components have the unique nature that they can be put in any of the available content categories. In order for this to work with those using web components, tags that are not, by default, part of the HTML spec (and therefore recognized as phrasing, flow, or another type) should be ignored.

I added the following test to https://github.com/kangax/html-minifier/blob/gh-pages/tests/minifier.js#L1191 and it failed:

```
  test('phrasing content with Web Components', function() {
    input = '<span><phrasing-element></phrasing-element></span>';
    output = '<span><phrasing-element></phrasing-element></span>';
    equal(minify(input, { html5: true }), output);
  });
```

pavax:
+1

mathieu-clement:
+1

mathieu-clement:
If I look at the code in htmlparser.js I can see that tags such as h3 are listed as phrasing content. This however is not what the HTML specification says:
https://html.spec.whatwg.org/multipage/dom.html#phrasing-content-2

You can see that h3 is not in this list. It is listed as "flow content".

alexlamsl:
[No they are not](https://github.com/kangax/html-minifier/blob/gh-pages/src/htmlparser.js#L91).

I think you may be confusing the tags that can contain [phrasing content _only_](https://html.spec.whatwg.org/multipage/semantics.html#the-h1,-h2,-h3,-h4,-h5,-and-h6-elements).

alexlamsl:
Fixed by #558

tiktaktok:
I'm confused, because h1-h6 are listed here:
https://html.spec.whatwg.org/multipage/dom.html#flow-content

PR #558, opened 2016-03-16
Handle phrasing content only for known tags

jshcrowthe:
Solves #557


dist/htmlminifier.js
```diff
@@ -95,6 +95,12 @@
...
 
+  var isKnownTag = function(tagName) {
+    // List comes from https://html.spec.whatwg.org/multipage/indices.html#elements-3
+    var knownTags = ["a","abbr","address","area","article","aside","audio","b","base","bdi","bdo","blockquote","body","br","button","canvas","caption","cite","code","col","colgroup","data","datalist","dd","del","details","dfn","dialog","div","dl","dt","em","embed","fieldset","figcaption","figure","footer","form","h1","head","header","hgroup","hr","html","i","iframe","img","input","ins","kbd","keygen","label","legend","li","link","main","map","mark","math","menu","menuitem","meta","meter","nav","noscript","object","ol","optgroup","option","output","p","param","picture","pre","progress","q","rp","rt","ruby","s","samp","script","section","select","small","source","source","span","strong","style","sub","summary","sup","svg","table","tbody","td","template","textarea","tfoot","th","thead","time","title","tr","track","u","ul","var","video","wbr"];
```
alexlamsl: Is this the same as `htmlTags` from #514?

jshcrowthe: The value of `htmlTags` in #514 includes several tags which are listed as obsolete (https://www.w3.org/TR/html5/obsolete.html#obsolete).

The value of `knownTags` (which, looking at the codestyle, I could refactor to use the `makeMap` fxn) is based off of the current supported list listed here: https://html.spec.whatwg.org/multipage/indices.html#elements-3

So they aren't the same, but with this specific function call already restricting the scope of the function to HTML 5 tags it shouldn't pose an issue.

alexlamsl: Okay, since this feature is guarded by `html5` flag, I guess we are safe to just stick to that spec by the letter.

`makeMap` gives better performance than `[].indexOf()`, so yes please use that :wink:

jshcrowthe: Can do, I will get that pushed first thing tomorrow morning. Thanks for taking a look at this!

jshcrowthe:
@alexlamsl: I have refactored the `[].indexOf()` usage. Let me know what you think!

alexlamsl:
@jshcrowthe LGTM :+1:

jshcrowthe:
Anything else I need to do?

alexlamsl:
Nothing I know of - just waiting for @kangax to parse this :wink:

dist/htmlminifier.js
```diff
@@ -95,6 +95,9 @@
...
   var phrasing = makeMap('a,abbr,area,audio,b,bdi,bdo,br,button,canvas,cite,code,data,datalist,del,dfn,em,embed,i,iframe,img,input,ins,kbd,keygen,label,link,main,map,mark,math,menu,meter,nav,noscript,object,ol,output,p,picture,pre,progress,q,ruby,s,samp,script,section,select,small,span,strong,sub,sup,svg,table,template,textarea,time,u,ul,var,video,wbr');
 
+  // List comes from https://html.spec.whatwg.org/multipage/indices.html#elements-3
+  var html5Tags = makeMap('a,abbr,address,area,article,aside,audio,b,base,bdi,bdo,blockquote,body,br,button,canvas,caption,cite,code,col,colgroup,data,datalist,dd,del,details,dfn,dialog,div,dl,dt,em,embed,fieldset,figcaption,figure,footer,form,h1,head,header,hgroup,hr,html,i,iframe,img,input,ins,kbd,keygen,label,legend,li,link,main,map,mark,math,menu,menuitem,meta,meter,nav,noscript,object,ol,optgroup,option,output,p,param,picture,pre,progress,q,rp,rt,ruby,s,samp,script,section,select,small,source,source,span,strong,style,sub,summary,sup,svg,table,tbody,td,template,textarea,tfoot,th,thead,time,title,tr,track,u,ul,var,video,wbr');
+
```
kangax, line 100: We should probably refactor all these at some point but LGTM for now.

jshcrowthe:
Awesome, thanks a bunch! Can we get a bugfix release for this?

dist/htmlminifier.js
alexlamsl, line 100: Yeah I had some internal back-and-forth on this but couldn't think of the best way. Any suggestions?

kangax, line 100: Something like `abbr: { content: 'phrasing', ... : ... }` would remove duplication but will make parsing slower...

PR #564, opened 2016-03-17
Throw Error not string

JohnNilsson:
Got bitten by the silent treatment from html-loader (another fix for the same issue https://github.com/webpack/html-loader/pull/57 )

### DIFF
--- a/src/htmlparser.js
+++ b/src/htmlparser.js
@@ -243,7 +243,7 @@
       }
 
       if ( html === last ) {
-        throw 'Parse Error: ' + html;
+        throw new Error('Parse Error: ' + html);
       }
     }
 


alexlamsl:
LGTM.

@XhmikosR I had to disable [no-throw-literal](http://eslint.org/docs/rules/no-throw-literal) when we migrate to `eslint` as well, so may be we can put that back on :sunglasses:

XhmikosR:
@alexlamsl: you shouldn't have done that, because the migration was a good reason to spot issues like this :P

Anyway, if you are up to it, try to make the ESLint configuration stricter.

alexlamsl:
@XhmikosR guilty as charged :disappointed: 

In my defense - when I was presented with 100+ errors in `htmlparser.js` alone after `eslint --init`, I thought I'll take the easy way out :sweat_drops: 

Let's sort out all the new bug reports from `1.3.0`, then I'll enable the rules alongside the source fixes and put them up for review :wink:

PR #568, opened 2016-03-18
fix the static analysis of Webpack

oliviertassinari:
That's is following this [short discussion](https://github.com/kangax/html-minifier/pull/531/files/0daf110770393f079da6d6fb66576fdcc6eafd35#r56428013).
Webpack is doing a static analysis of the source files do determine the used dependencies.
Using a runtime resolution of the dependencies isn't something he can handle out of the box:

``` js
require(path);
```

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -642,17 +642,13 @@
     return text;
   }
 
-  function getModule(name, path) {
-    // try to get global reference first
-    var result = global[name];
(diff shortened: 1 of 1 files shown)


alexlamsl:
@oliviertassinari thanks for looking into this :+1: 

@kangax I wonder if we should just make all `src/*.js` use plain `require()`, then have the `dist` build using `browserify`? That way we can avoid this `webpack` issue, plus we can drop/merge all those custom module builds in `assets`...

PR #569, opened 2016-03-19
consolidate global and require() usages

alexlamsl:
Hopefully this will keep the source code clean from multiple module import paths, and makes it more readable plus happy with things like `webpack` etc.


alexlamsl:
It's commit like this where I wish Github provides a diff mode that ignore whitespace changes... :sweat_smile:

alexlamsl:
So this will hopefully solve #568 and more - @oliviertassinari if you can verify this that would be helpful!

And we can actually review the [diff with whitespace changes ignored](https://github.com/kangax/html-minifier/pull/569/files?w=1) :clap:

assets/master.js
```diff
@@ -1,6 +1,6 @@
-(function() {
-  'use strict';
+'use strict';
```
XhmikosR: This is a browser file and strict is fine in the function.

index.html
```diff
@@ -4,10 +4,6 @@
     <meta charset="utf-8">
     <title>HTML minifier</title>
     <link rel="stylesheet" href="assets/master.css">
-    <script src="assets/cleancss-browser.js"></script>
-    <script src="assets/ncname-browser.js"></script>
-    <script src="assets/relateurl-browser.js"></script>
-    <script src="assets/uglify-browser.js"></script>
     <script src="dist/htmlminifier.min.js"></script>
```
XhmikosR: Can't we move this before the end body tag?

assets/master.js
alexlamsl: `strict` rule from `eslint` complaints - do you want me to add `/* eslint-disable */` instead?

alexlamsl:
@XhmikosR both comments addressed, PR rebased

assets/.eslintrc.json
```diff
@@ -1,5 +1,8 @@
 {
   "env": {
     "browser": true
+  },
+  "rules": {
+    "strict": "off"
```
XhmikosR: Have you just tried setting node to false without turning off strict?

index.html
```diff
@@ -266,6 +265,7 @@
       </div>
 
     </div>
+    <script src="dist/htmlminifier.min.js"></script>
```
XhmikosR: Maybe squash this commit with https://github.com/kangax/html-minifier/pull/569/commits/9f37bb689faef3501e6a5ae5b388047110c45017?

assets/.eslintrc.json
alexlamsl: No dice :confused: 

```
{
  "env": {
    "browser": true,
    "node": false
  }
}
```

gives

```
  114:2   error  Use the function form of 'use strict'  strict
```

alexlamsl:
@XhmikosR didn't know you can rearrange commits with `git rebase -i` before :open_mouth: 

Commits squashed.

assets/.eslintrc.json
XhmikosR: Weird, I'd expect it to handle the browser case. How about `strict: [2, "function"]`? Can you test if it works OK for both cases? Otherwise there is the `safe` option too.

alexlamsl: I think it's more to do with the Google Analytics code at the bottom of the script?

XhmikosR: If that's the case please just add eslint ignore around the snippet.

alexlamsl: `"node": false` would mean I need to add `/* global: require */` as well - but okay, I'll get them :sweat_smile:

XhmikosR: Then keep node...

alexlamsl: Alright:
- keep `node: true`
- change to `strict: [2, "function"]`
- add `/* eslint-disable */` before GA

Sounds about right?

XhmikosR: :+1:

alexlamsl:
@XhmikosR amended last commit to adjust eslint rules

alexlamsl:
Editted commit message to close relevant issue.

XhmikosR:
Ship it.

genintho:
@alexlamsl 

> It's commit like this where I wish Github provides a diff mode that ignore whitespace changes... :sweat_smile:

https://github.com/kangax/html-minifier/pull/569/files?w=1

Sadly you can not comment in this mode.

kangax:
Long time coming and is much needed by now! :)

kangax:
Next step — ES6 via babel :)

XhmikosR:
Will you make the switch? :)

BTW I'm thinking we might need to find another parser.

alexlamsl:
@genintho yeah, thanks for the tip :wink: 

@kangax is that the same babel that broke and went onto the news a few days back?

@XhmikosR thanks for the review! To switch HTML parser we'll need to decouple the logical dependencies first, but it's something to think about nonetheless.

kangax:
> @kangax is that the same babel that broke and went onto the news a few days back?

Nope, nothing to do with Babel. Babel broke because the author of one of the (npm) dependencies pulled them out.

PR #571, opened 2016-03-21
treat empty inline tags as text

alexlamsl:
@XhmikosR I've tested with https://mpc-hc.org/ as input with config from https://github.com/mpc-hc/mpc-hc.org/blob/master/Gruntfile.js - apart from the IE conditional comments the rest are now identical with this PR.


XhmikosR:
:+1:

XhmikosR:
So @alexlamsl @kangax when can we have a patch release?

alexlamsl:
@XhmikosR I've prepared #576 - so whenever @kangax is ready :wink:

PR #573, opened 2016-03-21
populate eslint with common rules

alexlamsl:
From https://github.com/kangax/html-minifier/commit/560e109e7f3fba6882b08c9d760404fc28754362#commitcomment-16783499

@XhmikosR I've incorporated the rules from https://github.com/giakki/uncss/blob/deps/.eslintrc - the omitted ones are almost all due to being included in the `recommended` set.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,6 +4,7 @@
   },
   "extends": "eslint:recommended",
   "rules": {
+    "block-scoped-var": "error",
     "brace-style": [
       "error",
       "stroustrup",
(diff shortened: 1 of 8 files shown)


alexlamsl:
Almost forgot - I've left `strict` out since I couldn't get the declaration style configured. Let's add that rule after #569

dist/htmlminifier.js
```diff
@@ -1698,7 +1698,7 @@
     return minify(value, options);
   };
 
-}(typeof exports === 'undefined' ? this : exports));
+})(typeof exports === 'undefined' ? this : exports);
```
XhmikosR, line 1701: We can go with `outside` instead.

.eslintrc.json
```diff
@@ -11,7 +12,88 @@
...
+    "comma-spacing": "error",
+    "comma-style": [
+      "error",
+      "last"
+    ],
+    "curly": "error",
+    "eol-last": "error",
+    "eqeqeq": [
+      "error",
+      "smart"
```
XhmikosR: I wouldn't go with smart.

XhmikosR:
Are you sure none of the added rules are not already covered by the eslint recommended?

tests/minifier.js
```diff
@@ -777,21 +777,21 @@
...
-             '<input type=\"text\">\r\n<textarea><\/textarea>\r\n<pre><\/pre>';
-    output = '<script type="text/javascript">var = "hello";</script>'                                +
-             '<style type="text/css">#foo { color: red;        }</style>'                            +
-             '<div><div><div>'                                                                       +
-             '<!-- hello --><div><!--! hello --><div><div class="">'                                 +
-             '<textarea disabled="disabled">     this is a textarea </textarea>'                     +
-             '</div></div></div></div></div></div>'                                                  +
-             '<pre>       \r\nxxxx</pre><span>x</span> <span>Hello</span> <b>billy</b> '             +
+    input = '<script type=\"text\/javascript\">var = \"hello\";<\/script>\r\n\r\n\r\n' +
+            '<style type=\"text\/css\">#foo { color: red;        }          <\/style>\r\n\r\n\r\n' +
```
XhmikosR, line 781: Maybe remove or align the lines before +.

alexlamsl:
Ah, I should mention this - I look up http://eslint.org/docs/rules/ as I scan through your list. So it's a matter of skipping the ones with a big tick next to it.

tests/minifier.js
alexlamsl, line 781: Sorry I don't quite follow what you mean...?

.eslintrc.json
alexlamsl: There's `== null` in `cli.js` and I'm not sure if I can get rid of that safely...

dist/htmlminifier.js
alexlamsl, line 1701: I'll have to edit some files either way `htmlparser.js` is `inline` - but `outside` it is then :wink:

.eslintrc.json
XhmikosR: Just add an ignore for that file then.

dist/htmlminifier.js
XhmikosR, line 1701: I don't mind which one, I just thought we were using outside already. So, just keep it as you have it here.

tests/minifier.js
XhmikosR, line 781: It was aligned before, no big deal.

alexlamsl, line 781: I have to change these due to `no-multi-spaces` complaining - and it's the only place in the project that rule has a trouble with...

alexlamsl:
@XhmikosR addressed your comments except those `+` lines inside `tests/minifier.js` - anything else?

XhmikosR:
Looks good to me. I'd just squash the patches and merge.

alexlamsl:
Will do~ :tada:

XhmikosR:
Maybe keep `inside`? It was used in more places I believe.

We should also ignore the Google Analytics snippet.

alexlamsl:
Oh alright then - Google Analytics is already `inside`, so that's even better :sunglasses:

alexlamsl:
Great timing - my `git push -f` failed due to Github throwing Angry Unicorns at me...

PR #587, opened 2016-03-26
add bootstrap CSS to benchmarks

alexlamsl:
@XhmikosR as per request :wink: 

In this particular benchmark we are far behind [Will Peavy](http://www.willpeavy.com/minifier/) - 269KB vs 233KB.


XhmikosR:
@alexlamsl: what options did you use? The only reason I can think of is will peavy having a bug and breaking the `<pre>` code.

alexlamsl:
@XhmikosR `benchmark.js` uses `sample-cli-config-file.conf` as options.

To begin with, they took out one of the two chunks of IE Conditional Comments.

Further inspecting the result I see snippets like `</pre>0</figure>`, `</pre>1</figure>`, ... 130 of them in total, which cannot be found in the original document. Chunks of markups may be getting uber-compressed into these magic numbers :wink:

XhmikosR:
Yeah, told you, it's the most logical explanation since it has many code snippets.

alexlamsl:
Sometimes I wish you or @kangax would just let me output an empty string - would reduce a lot of code complexity and makes the whole project so much more maintainable :ghost: 

I even have this idea of achieving larger-than-100% compression by randomly removing local files every time `cli.js` is run, but I guess that's probably over-kill...

alexlamsl:
Jokes aside - anything else you want me to do on this task @XhmikosR ?

XhmikosR:
Looks good to me, hoping for a completely automated solution in the future which will grab the pages each time :)

alexlamsl:
That and auto-update `README.md` after each run :sunglasses:

PR #590, opened 2016-03-29
Custom JS minifier for es6 (or coffee, or whatever)

duncanwerner:
I need to support es6 templates (backtick strings).  By default uglify-js doesn't support this, although there's a harmony branch.  I'd like to swap in a custom minifier.

Dropping in arbitrary minifiers is hard because the function calls might not match.  My solution is to add a "minifier" option to the minifyJS parameter, which is a function that can provide its own dependencies and call structure.  So the call looks like

``` js
htmlmin({
    collapseWhitespace: true,
    removeComments: true,
    minifyJS: {
        minifier: function( text, options ){
            var uglify = require( "uglify-js" );
            return uglify.minify(text, options).code;
        }
    },
    minifyCSS: true
});

```

and omitting the parameter leaves the default behavior.

v 1.3.1

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -3,8 +3,9 @@
 var CleanCSS = require('clean-css');
 var HTMLParser = require('./htmlparser').HTMLParser;
 var RelateUrl = require('relateurl');
-var UglifyJS = require('uglify-js');
 var utils = require('./utils');
+
+var defaultJSMinifier = "uglify-js";
 
 var log;
 if (typeof window !== 'undefined' && typeof console !== 'undefined' && typeof console.log === 'function') {
@@ -635,7 +636,8 @@ function minifyURLs(text, options) {
 
 function minifyJS(text, options) {
   try {
-    return UglifyJS.minify(text, options).code;
+    var Minifier = require( options.minifier ? options.minifier : defaultJSMinifier );
+    return Minifier.minify(text, options).code;
   }
   catch (err) {
     log(err);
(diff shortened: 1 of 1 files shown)


src/htmlminifier.js
```diff
@@ -635,7 +635,12 @@ function minifyURLs(text, options) {
 
 function minifyJS(text, options) {
   try {
-    return UglifyJS.minify(text, options).code;
+	if( options.minifier ){
+      return options.minifier.call( this, text, options );
```
alexlamsl, line 639: Why not just specify a `function` in `options.minifyJS`, then here just do:

``` js
if (typeof options === 'function') {
  return options(text);
} else {
  return UglifyJS.minify(text, options).code;
}
```

alexlamsl:
ES6 I can understand, but I don't think CoffeeScript would be using `<script type="text/javascript">`?

duncanwerner:
As far as I can tell, you're not checking the script type, although that might be implicit. If type checking were supported I'd suggest switching calls based on types, but that's a slightly more complex modification.

Regarding just using a function instead of options, that makes sense to me.

alexlamsl:
There are two places `minifyJS` is used:
- [event attributes](https://github.com/kangax/html-minifier/blob/gh-pages/src/htmlminifier.js#L264)
- [script tags](https://github.com/kangax/html-minifier/blob/gh-pages/src/htmlminifier.js#L981)

The latter case has [checks for script type](https://github.com/kangax/html-minifier/blob/gh-pages/src/htmlminifier.js#L190).

If you are extending `options.minifyJS`, you will also need to [modify the logic here](https://github.com/kangax/html-minifier/blob/gh-pages/src/htmlminifier.js#L608).

PR #594, opened 2016-04-01
sort attributes by frequency

alexlamsl:
This is a straight-forward implementation, with the following limitations/TODOs:
- [x] does not account for or propagate statistics to `processScripts`
- [x] does not omit removed attributes by other optimisations

PR #595, opened 2016-04-01
break up normalizeAttribute()

alexlamsl:
By-product of #594 - in trying things out I manage to break this 7-argument function into two.


alexlamsl:
@kangax IIRC it's a pet-peeve of yours, so PTAL :stuck_out_tongue_winking_eye:

src/htmlminifier.js
```diff
@@ -523,16 +519,31 @@ function normalizeAttribute(attr, attrs, tag, hasUnarySlash, index, options, isL
...
   }
 
+  return {
+    attr: attr,
+    name: attrName,
+    value: attrValue
+  };
+}
+
+function printAttr(normalized, hasUnarySlash, options, isLast) {
```
kangax: Maybe let's call it `buildAttr` instead? :)

alexlamsl: Done :wink:

PR #596, opened 2016-04-01
Feature remove chunk

ginnwork:
Added the ability to specify chunks of markup to delete. Useful for deleting development only scripts(ex: BrowserSync).

Input:

```
<p>Stays put</p>
<!-- htmlmin:remove -->
<script id="__bs__">
/* This script tag will be removed entirely */
</script>
<!-- htmlmin:remove-->
<script type="text/javascript" src="app/app.module.js"></script>
```

Output:

```
<p>Stays put</p>
<script type="text/javascript" src="app/app.module.js"></script>
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 /benchmarks/generated
 /node_modules/
 /npm-debug.log
+/nbproject/
--- a/test.js
+++ b/test.js
@@ -11,7 +11,8 @@ testrunner.run({
   code: './src/htmlminifier.js',
   tests: [
     './tests/lint.js',
-    './tests/minifier.js'
+    './tests/minifier.js',
+    './tests/remove.js'
   ],
   maxBlockDuration: 5000
 }, function(err, report) {
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -703,6 +703,10 @@ function minify(value, options, partialMarkup) {
     return token;
   });
 
+  // Remove specific comment chunks. Usefull to delete markup that isn't wanted
+  // in dist(ex: browsersync or livereload)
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -1,8 +1,9 @@
-*.sublime-project
-*.sublime-workspace
-.DS_Store
-/.jekyll-metadata
-/_site/
-/benchmarks/generated
-/node_modules/
-/npm-debug.log
```
alexlamsl, line 8: What happened here? Why are all the lines changed?

test.js
```diff
@@ -11,7 +11,8 @@ testrunner.run({
   code: './src/htmlminifier.js',
   tests: [
     './tests/lint.js',
-    './tests/minifier.js'
+    './tests/minifier.js',
+    './tests/remove.js'
```
alexlamsl, line 15: You need to make sure the new tests will run in the web version as well.

alexlamsl:
Overall I'm not sure whether the feature makes sense to be included in `html-minifier`. Unlike `htmlmin:ignore` which is required to workaround sections of the input which would otherwise be invalid HTML, this feature doesn't even relate to HTML at all, and the one-liner can be run before feeding the input into `minify()` without any loss of functionality.

.gitignore
ginnwork, line 8: Sorry, didn't mean to commit that. I'm new to the whole pull requests thing so I'm making some mistakes.

alexlamsl, line 8: No worries, I'm pretty new to this as well :ghost:

I am genuinely confused as to why it flags all the lines as changed even though most of them are visually identical.

PR #602, opened 2016-04-05
use commander instead of cli

alexlamsl:
Reason for migration:
- less dependencies
- shares with clean-css
- 2.5M vs 18M downloads per month
- 353 vs 288 lines for `cli.js`

Also fixes a bug with minifier options mutating when processing directory as input - ~~there is a performance penalty for this fix, which I'm looking into.~~ the performance difference is due to the bug resulting in some options being suppressed.


kangax:
LGTM (although no experience with either)

PR #604, opened 2016-04-05
maintain stable sort order

alexlamsl:
So further testing reveals that `benchmarks/amazon.html` gives different minified sizes from run to run, and this is due to the unique IDs generated by `ignoreCustomFragments` interfering with `sortClassName`.

Same bug would affect `<!-- htmlmin:ignore -->` and `sortAttributes` as well - this PR covers them all.


dist/htmlminifier.js
```diff
@@ -32566,17 +32566,28 @@ function createSortFns(value, options) {
     });
   }
 
+  function skipUID(token, uid) {
```
kangax, line 32569: I usually name boolean-returning functions as shouldSmth, canSmth, isSmth, etc.

kangax, line 32569: So let's rename to `shouldSkip...` and LGTM after that! :)

alexlamsl, line 32569: Done :wink:

PR #605, opened 2016-04-05
fix parsing of attribute quotes

alexlamsl:
So `preventAttributesEscaping` exposes a few bugs left over from HTML parser's handling of attribute quotes.

This PR fixes those plus a few tests.


src/htmlparser.js
```diff
@@ -294,15 +294,29 @@ function HTMLParser(html, handler) {
         if (args[5] === '') { args[5] = undefined; }
       }
 
+      function populate(index) {
+        customAssign = args[index];
+        value = args[index + 1];
+        if (value !== undefined) {
```
kangax: Hm, I think the convention we follow with `undefined` checks is to do them via `typeof` (or just `!value` if it can never be falsy at that point)

alexlamsl: `typeof v === 'undefined'` is different from `v === undefined` in that the former allows `v` to not have been declared yet, whereas the latter will cause a runtime error.

So I only use the former when I need to cover that specific case, to avoid when I have a typo with `v` :sweat_smile:

Either way I'm not too fussed, so I'll change it :wink:

PR #610, opened 2016-04-07
Improve online minifier

alexlamsl:
- synchronise minifier options
- `master.js` now reads from DOM instead of keeping its own list of options
  - only need to update `index.html` going forward
- tweak flattr code to avoid UI blocking, especially when running `index.html` locally
- `HTMLLint.populate()` to clear element when no lint errors are found
  - currently errors from previous run remain
- `grunt dist` now updates version number of `index.html`


alexlamsl:
Forgot to mention - I've also removed that disclaimer about beta software, just because. :ghost:

kangax:
:+1:

PR #614, opened 2016-04-08
enforce more eslint rules

alexlamsl:
@XhmikosR as per https://github.com/kangax/html-minifier/commit/cd86c425bcb2de6b399efbd22887598099ef3f69#commitcomment-17021892, I've went through the list and enforce most of them.

Take your time to review and see what you want add/remove from here :stuck_out_tongue_winking_eye:


.eslintrc.json
```diff
@@ -40,28 +50,57 @@
       }
     ],
     "keyword-spacing": "error",
+    "linebreak-style": "error",
+    "max-statements-per-line": "error",
```
XhmikosR: Do we have one line statements? If not, I'd remove it.

.eslintrc.json
```diff
@@ -40,28 +50,57 @@
...
     "no-spaced-func": "error",
     "no-throw-literal": "error",
     "no-trailing-spaces": "error",
+    "no-undef-init": "error",
+    "no-unmodified-loop-condition": "error",
+    "no-unneeded-ternary": "error",
     "no-unused-expressions": "error",
+    "no-use-before-define": [
+      "error",
+      "nofunc"
```
XhmikosR, line 96: Why do you use `nofunc`? In theory we shouldn't have any undefs?

.eslintrc.json
```diff
@@ -70,6 +109,7 @@
       "error",
       "always"
     ],
+    "operator-linebreak": "error",
```
XhmikosR: Maybe add `after` even though it's default, it will be clear.

.eslintrc.json
alexlamsl: I was hoping to forbid future additions of multiple statements like `pos.x = x; pos.y = y;`

.eslintrc.json
alexlamsl, line 96: We have circular references of functions, so this make it only check for `var`

alexlamsl:
@XhmikosR rebased, addressed `operator-linebreak`, also added `no-undefined` which achieve the thing you want with `valid-typeof`

.eslintrc.json
alexlamsl: On closer inspection, the rule (which has `max=1` by default) does not complain about this in `htmlparser.js`:

```
if (args[3] === '') { delete args[3]; }
if (args[4] === '') { delete args[4]; }
if (args[5] === '') { delete args[5]; }
```

which according to [their documentation](http://eslint.org/docs/rules/max-statements-per-line#max) is clearly incorrect.

I'll go with your suggestion and remove this option then.

.eslintrc.json
alexlamsl, line 96: This is what happens without `nofunc`:

```
src/htmlminifier.js
   267:22  error  'minifyJS' was used before it was defined            no-use-before-define
   285:14  error  'minifyURLs' was used before it was defined          no-use-before-define
   298:14  error  'minifyCSS' was used before it was defined           no-use-before-define
   302:12  error  'isMetaViewport' was used before it was defined      no-use-before-define
   346:21  error  'minify' was used before it was defined              no-use-before-define
   354:14  error  'minify' was used before it was defined              no-use-before-define
   729:8   error  'minify' was used before it was defined              no-use-before-define
  1157:13  error  'joinResultSegments' was used before it was defined  no-use-before-define

src/htmlparser.js
   90:30  error  'joinSingleAttrAssigns' was used before it was defined  no-use-before-define
  168:42  error  'parseEndTag' was used before it was defined            no-use-before-define
  174:29  error  'parseStartTag' was used before it was defined          no-use-before-define
  177:11  error  'handleStartTag' was used before it was defined         no-use-before-define
  194:26  error  'parseStartTag' was used before it was defined          no-use-before-define
  232:7   error  'parseEndTag' was used before it was defined            no-use-before-define
  242:5   error  'parseEndTag' was used before it was defined            no-use-before-define
  271:7   error  'parseEndTag' was used before it was defined            no-use-before-define
  276:9   error  'parseEndTag' was used before it was defined            no-use-before-define
  281:7   error  'parseEndTag' was used before it was defined            no-use-before-define

? 18 problems (18 errors, 0 warnings)
```

XhmikosR, line 96: OK, just put it back then.

alexlamsl, line 96: So I think that's all 3 comments addressed then :tada:

PR #615, opened 2016-04-08
Cache variable in for loops.

XhmikosR:
/CC @alexlamsl @kangax: mostly for consistency since the performance improvement shouldn't be too big.


alexlamsl:
I'd normally hope for the VM to optimise enough, but this won't hurt too.

PR #617, opened 2016-04-09
Improve benchmarks

alexlamsl:
- benchmark files are now downloaded instead of storing offline copies
- HTMLCompressor will now throw 500 for Amazon, so need to workaround that

TODO output results to `README.md`


benchmark.js
```diff
@@ -328,6 +331,35 @@ run(fileNames.map(function(fileName) {
...
       done();
     });
+  }
+
+  function get(site, callback) {
+    var options = url.parse(site);
+    var protocol = options.protocol === 'https:' ? https : http;
+    protocol.get(options, function(res) {
+      var status = res.statusCode;
+      if (200 <= status && status < 300) {
```
XhmikosR: `yoda` :p

PR #621, opened 2016-04-12
sychronise qunit version for node and web

alexlamsl:
Previously we use `qunit` which is locked to QUnit 1.10.0, whereas on the web we always use the latest version (currently 1.23.0).

This PR switches to `qunitjs` which is more direct, and have `grunt` make sure the web tests uses the same QUnit version as the node tests as well.

Also removing unnecessary dependencies - managed to cut down by 10 modules.


alexlamsl:
Also put `eslint` checks before `dist` inside `test` so it'd fail earlier.

PR #625, opened 2016-04-14
minify URLs within CSS

alexlamsl:
After thinking about this some more, I think we can safely turn on this optimisation for inline CSS.

Also fixes some typo and styles in tests.

### DIFF
--- a/tests/minifier.js
+++ b/tests/minifier.js
@@ -535,34 +535,34 @@ test('remove CDATA sections from scripts/styles', function() {
   output = '<script>alert(11)</script>';
   equal(minify(input, { minifyJS: true }), output);
 
-  input = '<style><![CDATA[\np.a{backgourd:red}\n]]></style>';
+  input = '<style><![CDATA[\np.a{background:red}\n]]></style>';
   equal(minify(input), input);
-  output = '<style><![CDATA[ p.a{backgourd:red}\n]]></style>';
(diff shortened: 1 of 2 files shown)


tests/minifier.js
```diff
@@ -535,34 +535,34 @@ test('remove CDATA sections from scripts/styles', function() {
   output = '<script>alert(11)</script>';
   equal(minify(input, { minifyJS: true }), output);
 
-  input = '<style><![CDATA[\np.a{backgourd:red}\n]]></style>';
```
kangax, line 538: lol

alexlamsl, line 538: If it weren't for VS Code suggesting keywords harvested from the file, I wouldn't have spotted this :disappointed_relieved:

kangax, line 538: Nice. How are you liking VS Code btw? I've been using Atom recently (after having used Sublime for few years) but hear good things about VSC too.

alexlamsl, line 538: It's pretty lightweight and gets the job done, with pretty good hints / auto-completion, integrates well with git and won't get in my way like a full-featured IDE would.

So yeah I've been liking it so far :+1: 

My history of code editors:
- `edit.com`
- Notepad
- Visual Studio
- Eclipse
- NetBeans
- VS Code

PR #633, opened 2016-04-16
remove lint functionalities

alexlamsl:
So this together with the work on https://github.com/kangax/html-lint should take care of the separation.

@XhmikosR I have to include `dist/*.js` in this changeset, since otherwise the online minifier will break...


alexlamsl:
@kangax @XhmikosR question - after this, do we bump major or just minor version?

kangax:
Let's do major to play it safe.

XhmikosR:
Yeah agreed.

PR #640, opened 2016-04-19
Custom processor for minify{CSS,JS,URLs}

alexlamsl:
Allow functions for `minify{CSS,JS,URLs}` which will override the default minifiers we use.

The only change which affects existing use cases is the removal of trailing semi-colons in event attributes. They were trimmed regardless of `minifyJS` before, which is now handled by `minifyJS` and unified with same logic for script blocks.


alexlamsl:
@kangax would like to see if you think this is a good idea or not

PR #684, opened 2016-07-04
Remove unwanted element attributes by name

### LINKED ISSUE #322: is it possible to remove style attributes ?
i was hoping to do something like:

``` js
minify(HTMLSTRING,{
    attrBlackList: ['style']
});
```

is it possible to achieve the same functionality ?

HitkoDev:
Add an option to blacklist attributes by matching them against an array of values or regular expressions. Closes #322.


mdreizin:
@kangax did you have any chance to review it?

I also have some cases when I ned to remove some attributes (for example `data-test="xxx"`).

CedarDCX:
Any idea when this feature might get released? It would be a great additional feature which would allow us to remove redundant data-react-* attributes from our markup.

PR #700, opened 2016-07-30
upgrade devDependencies

alexlamsl:
grunt-contrib-uglify 2.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
   "devDependencies": {
     "grunt": "1.0.x",
     "grunt-browserify": "5.0.x",
-    "grunt-contrib-uglify": "1.0.x",
+    "grunt-contrib-uglify": "2.0.x",
     "grunt-eslint": "19.0.x",
     "phantomjs-prebuilt": "2.1.x",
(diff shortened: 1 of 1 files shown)


XhmikosR:
Note that screwIE8 is enabled by default in v2.0.0. So if you care about IE8 you need to set it to false.

alexlamsl:
Thanks for the reminder - I don't think this codebase has IE8 compatibility, but I'll try and double check with a WinXP VM later this evening.

XhmikosR:
There's nothing to check :p

If you want IE < 9 compatibility just set the option to false.

alexlamsl:
My point was if the pre-minified code already using features not supported by IE8, than `screwIE8` or not is a bit moot...

XhmikosR:
Yup, it already fails on IE8 so I guess just ignore it.

```
SCRIPT438: Object doesn't support property or method 'keys' 
htmlminifier.min.js, line 7 character 2382
SCRIPT5007: The value of the property 'require' is null or undefined, not a Function object 
master.js, line 4 character 3
```

PR #710, opened 2016-09-06
Add <br> to list of self-closing inline tags.

mrcsmcln:


alexlamsl:
Changes don't pass the unit tests, please take a look.

Would you mind elaborating on why you want to apply this change, and what issues are you aiming to fix? Adding tests to clarify your intention would also be helpful.

alexlamsl:
Invalid changes - closing.

PR #714, opened 2016-09-14
Respect line breaks around custom fragments

gzzhanghao:
It leads to an issue with following html fragment:

``` html
<script>
// some comment
<%customFragment%>
</script>
```

After replacing `customFragments`, it becomes

``` html
<script>
// some comment    (ignored id)
</script>
```

And the custom fragment will be removed in the minifyJS stage.


alexlamsl:
Can you add tests relevant to the use cases you are trying to fix?

gzzhanghao:
Ehm... after digging deeper I found a `preserveLineBreaks` option which can solve this issue. Thanks for your reply anyway ;)
BTW, I think the `preserveLineBreaks` option should be true by default, it won't cost more spaces and can avoid lots of potential issues.

alexlamsl:
@gzzhanghao `preserveLineBreaks` would cost more for cases when all whitespaces can be safely removed. In either case, I'm glad you have figured out a solution/workaround to this problem, as it involves `uglify-js` which does complicate matters.

PR #716, opened 2016-09-18
add customFragmentsInScripts option

gzzhanghao:
It leads to an issue with following html fragment:

``` html
<script>
function a() {
  b()
  <?php if (false): ?>
    c()
  <?php endif; ?>
}
</script>
```

After minified, it becomes:

``` html
<script>function a(){b(),}</script>
```

The trailing comma leads to a syntax error.

The only method I found is to prevent minifying javascripts when there's a customFragment inside the script tag.

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -865,6 +865,12 @@ function minify(value, options, partialMarkup) {
     });
   }
 
+  function restoreCustomFragments(text) {
+    return text.replace(uidPattern, function(match, prefix, index) {
+      return ignoredCustomMarkupChunks[+index][0];
+    });
+  }
+
   var customFragments = options.ignoreCustomFragments.map(function(re) {
     return re.source;
   });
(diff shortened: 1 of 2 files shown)


alexlamsl:
Can this not be implement with `minifyJS` by the use of [`function(text, inline){}`](https://github.com/kangax/html-minifier/blob/v3.0.3/tests/minifier.js#L730-L746)?

gzzhanghao:
@alexlamsl I believe not :(

Custom fragments were replaced by the placeholders at that moment, so you can't easily find out whether there's a customFragment inside the script.

Setting `sequence: false` for minifyJS may help in this issue, but I think it's better to have such an option to prevent other potential issues.

alexlamsl:
Right you are. :+1: 

I think rather than introducing this flag, if we can literally make this custom fragment case works by default that would be better. So if this option you've introduced work without any regressions, we might as well have this behaviour by default.

Failing that, I still prefer offloading this to a custom `minifyJS` processor if possible, if not because we can employ the same tricks for `minifyCSS`. After all, `uglify-js` and `clean-css` do not have the same sort of `ignoreCustomFragments` functionality as `html-minifier` does.

Would this work if I were to restore all the substitutions before passing the script string to `function(text, inline) {...}`?

alexlamsl:
So coming back to this. One way to think of it is that `uglify-js` and `clean-css` should be fully responsible for the content, so we should just restore any custom fragment substitutions and pass the origin string directly at them. This however would break a lot of existing cases where the current logic does work properly, and many of which produce good savings.

So I think the flag you've introduced is probably quite a good solution to this issue. A few things:
- the flag currently only takes care of JavaScript, but we do have similar issues with CSS as well
- one option is to make two flags, one for JS and one for CSS
- can we have the flag the other way round, i.e. set to `true` to pass custom fragments as-is?
- please exclude `dist/htmlminifier.js` from the commit

anonyco (approved):

PR #719, opened 2016-09-21
respect line breaks around custom fragments

gzzhanghao:
This PR aims to fix the same issue with #714 

The issue still exists after setting `preserveLineBreaks` option.

``` html
<script>//
<% ... %>
</script>
```

Expected:

``` html
<script>
<% ... %>
</script>
```

Actual:

``` html
<script></script>
```

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -888,7 +888,9 @@ function minify(value, options, partialMarkup) {
       }
       var token = uidAttr + ignoredCustomMarkupChunks.length;
       ignoredCustomMarkupChunks.push(match);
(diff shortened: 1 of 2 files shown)


alexlamsl:
Looking at that regex, cases like `<script>\n\t<% ... %></script>` might not be covered...?

Edit: never mind, I see how this can work just fine.

alexlamsl:
As stated in #716, I think we should just aim to tackle the larger issue of having custom fragments in all sorts of locations within scripts and stylesheets, either by default or using custom processors.

gzzhanghao:
Fixed in #720

PR #720, opened 2016-09-21
handle line-breaks around custom fragments gracefully

alexlamsl:
@gzzhanghao this is what I have in mind with regards to #716 and #719. Does this address all your cases?

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -858,36 +858,38 @@ function minify(value, options, partialMarkup) {
     return token;
   });
 
+  function escapeFragments(text) {
+    return text.replace(uidPattern, function(match, prefix, index) {
+      var chunks = ignoredCustomMarkupChunks[+index];
+      return chunks[1] + uidAttr + index + chunks[2];
+    });
+  }
+
   var customFragments = options.ignoreCustomFragments.map(function(re) {
     return re.source;
   });
   if (customFragments.length) {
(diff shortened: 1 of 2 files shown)


gzzhanghao:
It works like a charm! Can't wait to merge this :smiley:

gzzhanghao:
The issue in #716 still exists in this branch. I think we should restore all substitutions in the `escapeFragments` method.

alexlamsl:
@gzzhanghao thanks for the review - I'll go fix it up now... :wink:

alexlamsl:
Just tried your example in #716 and it seems to work:

```
> minify(`<script>
... function a() {
.....   b()
.....   <?php if (false): ?>
.....     c()
.....   <?php endif; ?>
..... }
... </script>`, {minifyJS:true})
'<script>function a(){b(),\n<?php if (false): ?>\n,c(),\n<?php endif; ?>\n}</script>'
```

What other options have you specified which produce the incorrect results as you described?

gzzhanghao:
@alexlamsl It becomes `<script>function a(){b(),\n\n}</script>` after running in php runtime. The trailing comma leads to a syntax error :(

Custom fragments are not valid javascript stuff, I think we can't simply treat it as a identifier in javascript.

alexlamsl:
@gzzhanghao now I see what you mean.

Yeah, this one is inherently tricky, and the same issue would arise with `minifyCSS` as well. I guess I'll merge this for now and think about #716 some more, as they are separate issues...

PR #729, opened 2016-10-06
Add clean-css advanced optimisations note

simontabor:
Advanced optimisation are disabled in clean-css for some reason (should they be re-enabled?)

This left me confused for quite some time until I actually inspected the source of html-minifier (I'd expect it to use the clean-css defaults).


alexlamsl:
That is odd indeed. I have traced it back to https://github.com/kangax/html-minifier/commit/2c2217c4b9af3a2848fca0acb48f23920ae6dcf0 so far, but the flag has been manipulated before that as well.

Given that `clean-css` has updated since, perhaps we should investigate whether the defaults are good enough these days and remove that `advanced=false` instead.

/cc @XhmikosR

XhmikosR:
I don't recall doing this on purpose, but sounds like something I'd do. The reason for this is pretty simple. clean-css's advanced optimizations break stuff numerous times. So I went with the safe solution. And you shouldn't change this IMO.

alexlamsl:
@XhmikosR thanks for the clarification. I shall take your advice rather than risking some nasty surprises then.

alexlamsl:
@simontabor thanks for the change.

simontabor:
thanks for the quick merge 👍

PR #736, opened 2016-11-03
Remove change-case from dependencies

vlyahovich:
Change-case lib is too distributed into packages that not needed and it has too much functionality twat we aren't using. Also we really need only one method in `cli.js` from it. So I suggest to remove this dependencies in favor of a simplified method.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -27,10 +27,11 @@
 
 'use strict';
 
-var changeCase = require('change-case');
+var camelCase = require('camel-case');
 var fs = require('fs');
 var info = require('./package.json');
 var minify = require('./' + info.main).minify;
+var paramCase = require('param-case');
 var path = require('path');
 var program = require('commander');
 
@@ -138,7 +139,7 @@ var mainOptions = {
 var mainOptionKeys = Object.keys(mainOptions);
 mainOptionKeys.forEach(function(key) {
   var option = mainOptions[key];
-  key = '--' + changeCase.paramCase(key);
+  key = '--' + paramCase(key);
   if (Array.isArray(option)) {
     var optional = option[1] === parseJSON;
     program.option(key + (optional ? ' [value]' : ' <value>'), option[0], option[1]);
@@ -197,7 +198,7 @@ program.arguments('[files...]').action(function(files) {
 function createOptions() {
   var options = {};
   mainOptionKeys.forEach(function(key) {
-    var param = program[changeCase.camelCase(key)];
+    var param = program[camelCase(key)];
     if (typeof param !== 'undefined') {
       options[key] = param;
     }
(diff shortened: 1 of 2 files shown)


vlyahovich:
Seems like `jslint` problems already in main branch

alexlamsl:
I don't see any reasons to remove that dependency which works as it is. I may be persuaded if we depend on the (two) individual sub-modules rather than `change-case` directly, but even then it's not exactly a big deal.

dist/htmlminifier.js
```diff
@@ -7,7 +7,7 @@
 require=(function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
 (function (process,__filename){
 /** vim: et:ts=4:sw=4:sts=4
- * @license amdefine 1.0.0 Copyright (c) 2011-2015, The Dojo Foundation All Rights Reserved.
+ * @license amdefine 1.0.1 Copyright (c) 2011-2016, The Dojo Foundation All Rights Reserved.
```
alexlamsl, line 10: Please don't commit files under `dist/` - they will be generated and updated when a release is made.

alexlamsl (changes requested):

vlyahovich:
Yes, I agree that it is not a big deal. But having many unnecessary packages adds up to build time. I've done some changes, can you please look one more time? Thanks!

cli.js
```diff
@@ -27,7 +27,8 @@
 
 'use strict';
 
-var changeCase = require('change-case');
+var camelCase = require('camel-case');
+var paramCase = require('param-case');
```
alexlamsl: Please sort these `require()` in alphabetical order

package.json
```diff
@@ -54,7 +54,8 @@
     "test": "grunt test"
   },
   "dependencies": {
-    "change-case": "3.0.x",
+    "camel-case": "3.0.x",
+    "param-case": "2.1.x",
```
alexlamsl: Please sort these dependencies in alphabetical order

alexlamsl (changes requested):
Looks okay with some final nits.

vlyahovich:
Done with review. Please take another look.

alexlamsl (approved):

alexlamsl:
Thanks for working on this.

vlyahovich:
Thanks! Appreciate your work.

PR #738, opened 2016-11-06
allow backreference usage in ignoreCustomFragments

### LINKED ISSUE #733: Regular Expression backreference issue in ignoreCustomFragments
Hi Juriy, 

The regexp backreferences usage in `ignoreCustomFragments` was affected from 3.0.3 version, here my case:

```
{
  ignoreCustomFragments: [ /<(WC@[\s\S]*?)>(.*?)<\/\1>/ ]
}
```

The grouping update of regexp expression in the code below (`htmlminifier.js`) was the problem because the `\1` backreference points to first group and not to `ignoreCustomFragments` group expression (so logically my regexp stop working):

in v3.0.3:
https://github.com/kangax/html-minifier/blob/89c29688d84b97de949b47dda66e72201760048a/src/htmlminifier.js#L865

> new RegExp('\s_(?:' + customFragments.join('|') + ')+\s_', 'g')

in v3.1.0:
https://github.com/kangax/html-minifier/blob/969d21a0a86696d0bb148e3c6bd572bddd9049c5/src/htmlminifier.js#L872

> new RegExp('(\s_)(?:' + customFragments.join('|') + ')+(\s_)', 'g');

any plans of update ? 
Thks !!!

alexlamsl:
fixes #733

### DIFF
--- a/tests/minifier.js
+++ b/tests/minifier.js
@@ -1813,6 +1813,18 @@ QUnit.test('Ignore custom fragments', function(assert) {
     collapseWhitespace: true,
     trimCustomFragments: true
   }), output);
+
+  input = 'foo <WC@bar> baz moo </WC@bar> loo';
+  assert.equal(minify(input, {
+    collapseWhitespace: true,
+    ignoreCustomFragments: [
+      /<(WC@[\s\S]*?)>(.*?)<\/\1>/
+    ]
+  }), input);
+  output = 'foo<wc @bar>baz moo</wc>loo';
+  assert.equal(minify(input, {
+    collapseWhitespace: true
+  }), output);
 });
 
 QUnit.test('bootstrap\'s span > button > span', function(assert) {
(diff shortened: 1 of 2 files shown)


tests/minifier.js
```diff
@@ -1813,6 +1813,18 @@ QUnit.test('Ignore custom fragments', function(assert) {
...
   }), output);
+
+  input = 'foo <WC@bar> baz moo </WC@bar> loo';
+  assert.equal(minify(input, {
+    collapseWhitespace: true,
+    ignoreCustomFragments: [
+      /<(WC@[\s\S]*?)>(.*?)<\/\1>/
+    ]
+  }), input);
+  output = 'foo<wc @bar>baz moo</wc>loo';
```
outaTiME, line 1824: The output must preserve both tags like input:

probably as follow:

```
output = 'foo<WC@bar>baz moo</WC@bar>loo';
```

alexlamsl: This is the case _without_ `ignoreCustomFragments` - the case you are interested in is a line above, when nothing's changed, i.e. expects to equal to `input` 😉

PR #737, opened 2016-11-06
fix no-useless-escape warnings from eslint 3.9.1

alexlamsl:


alexlamsl:
Updated version of `eslint` has stricter `no-useless-escape` checks for regular expressions, so fixing them in the source code.

Thanks to @vlyahovich for noticing this.

PR #742, opened 2016-11-10
Add ability to remove only empty `<script>` and `<style>` tags

sparanoid:
I'd like the ability to remove empty `<script>` and `<style>` tags even when `removeEmptyElements` is disabled, these empty tags are much more safer to be removed than other empty elements.

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -1113,6 +1113,22 @@ function minify(value, options, partialMarkup) {
           currentChars += '|';
         }
       }
+
+      if (options.removeEmptyScriptTags) {
+        if (isElementEmpty && tag === 'script') {
+          removeStartTag();
+        }
+        optionalStartTag = '';
+        optionalEndTag = '';
+      }
+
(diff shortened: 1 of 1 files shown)


alexlamsl (changes requested):
I'd prefer extending `removeEmptyElements` then adding two new options.

Also some tests for the new functionality would be nice.

anonyco:
Would this remove `<script src="/path/to/file.js" async defer></script>` because it has no content (only an `src` attribute)?

src/htmlminifier.js
```diff
@@ -1113,6 +1113,22 @@ function minify(value, options, partialMarkup) {
           currentChars += '|';
         }
       }
+
+      if (options.removeEmptyScriptTags) {
+        if (isElementEmpty && tag === 'script') {
```
anonyco, line 1118: I suggest changing this line to `        if (isElementEmpty && tag === 'script' && !attrs.src) {` so that external scripts do not get removed.

anonyco (approved):

Issue #396, opened 2015-08-14
minify css inside <style> tag


waltonryan:
I didn't notice any options to minify the css inside an inline style tag. I was hoping to use htmlmin for this purpose, in minifying email html templates where inline style is required. 

Issue is that it doesn't seem to minify any of the content inside a "style" tag. Is there an option for this? A quick fix?

Thanks.

kangax:
Pretty sure style contents should be minified. If not, it's a bug.

waltonryan:
Ok, then it probably is a bug. Here are my options. let me know if you would like any more info.

```
// package.json

...
"grunt-contrib-htmlmin": "^0.4.0",
...
```

```
// Gruntfile.js

...
    htmlmin: {
      build: {
        options: {
          removeComments: true,
          collapseWhitespace: true
        },
        files: [{
          expand: true,
          cwd: 'build/',
          src: ['*.html', 'templates/**/*.hbs'],
          dest: 'build/min/'
        }]
      }
    }
...
```

Output Snippet:

```
// min/example-email.html

<!DOCTYPE html><html><head><meta name="viewport" content="width=device-width"><meta http-equiv="Content-Type" content="text/html; charset=UTF-8"><title>Basic Email Example</title><style type="text/css">.ReadMsgBody {
width: 100%;
}
.ExternalClass {
width: 100%;
}
.ExternalClass {
line-height: 100%;
}
body {
-webkit-text-size-adjust: 100%; -ms-text-size-adjust: 100%;
}
...
```

You can see that the inline style sheet is not minified starting at the first '{'. Thanks for your help.

kangax:
Something else must be off because it works for me on https://kangax.github.io/html-minifier/

Tried with this:

``` html
<!DOCTYPE html><html><head><meta name="viewport" content="width=device-width"><meta http-equiv="Content-Type" content="text/html; charset=UTF-8"><title>Basic Email Example</title><style type="text/css">.ReadMsgBody {
width: 100%;
}
.ExternalClass {
width: 100%;
}
.ExternalClass {
line-height: 100%;
}
body {
-webkit-text-size-adjust: 100%; -ms-text-size-adjust: 100%;
}
</style>
</head>
<body></body></html>
```

waltonryan:
I'll take a closer look. Thanks.

waltonryan:
I found the issue. I needed to set the minifyCSS option to true.

```
    htmlmin: {
      build: {
        options: {
          ...
[...]

PR #759, opened 2016-11-22
Update README.md

revelt:
Explicitly mentioning the minification of the CSS in the `<head>` of the HTML to save people's time and to prevent false issues like #396 being raised. It is especially relevant to all email developers using this library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,10 @@ Output of resulting markup (e.g. `<p>foo</p>`)
 
 HTMLMinifier can't know that original markup was only half of the tree; it does its best to try to parse it as a full tree and it loses information about tree being malformed or partial in the beginning. As a result, it can't create a partial/malformed tree at the time of the output.
 
+### Minifying email templates
+
+By default, style tags within the `<head>` of the HTML are not minified. To do that, set `minifyCSS: true`. In theory, you can even minify _only_ the CSS styles in the `<head>` because it is separate from the HTML minification.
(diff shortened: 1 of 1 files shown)


alexlamsl:
`minifyCSS` is needed for _any_ CSS minification to take place, be it `<style>` in `<head>` or `<body>`, or even `style="..."` within elements.

So I failed to see how this additional paragraph would clarify the situation.

revelt:
Yes, but nobody puts linebreaks within inline CSS. CSS minification here is mainly for `<head>` styles. I found issue #396 by wanting to raise such issue myself. In my humble opinion it should be **on** by default because if I'm going to minify HTML, I'll want _everything_ minified.

It's your call to consider, whatever you feel is the most appropriate. Feel free to edit if necessary or amend completely 😀

alexlamsl:
`minifyCSS` does more than just removing line-breaks :sweat_smile: 

But yes, I agree with the general issue of not having a recommended set of options on by default seems to be catching a lot of people out. OTOH, this would be a major change, and I don't think we've arrived to a consensus on what those options should be yet.

May be adding a general warning of "all minification options are off by default" will incentivise people to actually read the options table? Either way, I'm afraid the text in this PR is too narrowly-scoped for the general audience, which would cause more confusion than clarification IMHO.

Thanks for the heads up nonetheless - I'll see what I can add before the next release.

PR #772, opened 2017-01-23
upgrade to clean-css 4.0

alexlamsl:
Removed workarounds put in place for previous versions of `clean-css`, namely HTML comment wrapper and manual disabling of `advanced` optimisations.

Even though the configurations for `clean-css` has been renamed and rearranged drastically, `html-minifier` with its default `minifyCSS: true` behaviour should be identical as before. So I am thinking of a minor version bump for `html-minifier`.

/cc @kangax @XhmikosR


XhmikosR:
I haven't looked v4. Do they have advanced optimizations disabled by default?

alexlamsl:
@XhmikosR that is indeed the case. `advanced` is now level 2 optimisations, and `clean-css` by default is on level 1

https://github.com/jakubpawlowicz/clean-css#optimization-levels

PR #791, opened 2017-03-10
clean-css and UglifyJS2 options

da0ka:
http://kangax.github.io/html-minifier
Above page doesn't have options for clean-css and UglifyJS2. Please add options like bellow.
https://jakubpawlowicz.github.io/clean-css
http://lisperator.net/uglifyjs


alexlamsl:
Not sure what you've done wrong with this PR, but you might want to open another one as it says there are 50 files changed and certainly doesn't look correct.

PR #807, opened 2017-05-06
fix `sortClassName` on custom fragments

### LINKED ISSUE #805: Quote matching issue in ignoreCustomFragments
Given this Handlebars template as input:

````html
<span class="{{classes}}"></span>
````

When `removeAttributeQuotes: true` is set, the quotes will (presumably rightfully) be removed giving this result:

````html
<span class={{classes}}></span>
````

To keep the quotes (because the variable can resolve to a value containing spaces) I added `ignoreCustomFragments: [/"\{\{.*?\}\}"/]`, which results in double double-quotes like in the result:

````html
<span class=""{{classes}}""></span>
````

Is that expected? It seems the value of `removeAttributeQuotes` has no effect on these quotes.

### PR DESCRIPTION
**Pull request by alexlamsl**
Prefer sub-optimal sort over extraneous whitespaces.

fixes #805

### DIFF
--- a/tests/minifier.js
+++ b/tests/minifier.js
@@ -3061,11 +3061,10 @@ QUnit.test('sort style classes', function(assert) {
     ignoreCustomFragments: [/<#[\s\S]*?#>/],
     sortClassName: false
   }), input);
-  output = '<div class="foo_bar nav_sv_fo_v_column <#=(j === 0) ? \'nav_sv_fo_v_first\' : \'\' #> "></div>';
   assert.equal(minify(input, {
     ignoreCustomFragments: [/<#[\s\S]*?#>/],
     sortClassName: true
-  }), output);
+  }), input);
 
   input = '<a class="0 1 2 3 4 5 6 7 8 9 a b c d e f g h i j k l m n o p q r s t u v w x y z"></a>';
   assert.equal(minify(input, { sortClassName: false }), input);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/minifier.js [hunk 1]
```diff
@@ -3075,6 +3074,14 @@ QUnit.test('sort style classes', function(assert) {
   assert.equal(minify(input, { sortClassName: false }), input);
   output = '<a class="add createSorter keys sort"></a>';
   assert.equal(minify(input, { sortClassName: true }), output);
+
+  input = '<span class="sprite sprite-{{sprite}}"></span>\n<span class="{{sprite}}"></span>';
+  output = '<span class="sprite sprite-{{sprite}}"></span> <span class="{{sprite}}"></span>';
+  assert.equal(minify(input, {
+    collapseWhitespace: true,
+    ignoreCustomFragments: [/{{.*}}/],
```
silverwind: While it doesn't really matter with this input, I prefer using non-greedy `.*?` so it does not match unintential parts when multiple `{{}}` expressions are on the same line, for example: https://regex101.com/r/M50E40/1

silverwind (approved):
LGTM, wasn't able to reproduce the issue anymore with a few variations of inputs.

````js
var minify = require('.').minify;
var opts = {
  collapseWhitespace: true,
  ignoreCustomFragments: [/{{.*?}}/],
  sortClassName: true,
};

console.log(minify(`<span class="sprite sprite-{{sprite}}"></span>`, opts));
console.log(minify(`<span class="{{sprite}}"></span>`, opts));
console.log(minify(`<span class="{{sprite}}-sprite sprite"></span>`, opts));
console.log(minify(`<span class="sprite-{{sprite}}-sprite"></span>`, opts));
console.log(minify(`<span class={{sprite}}></span>`, opts));
````

tests/minifier.js
alexlamsl: I'm well aware of that, just didn't worry too much as this is a single test case of specific input.

As you can see in https://github.com/kangax/html-minifier#options-quick-reference, the default values for `ignoreCustomFragments` is not only non-greedy, but can also handle multi-line values in between the template markers.

Nonetheless, I'll add the `?` to make it easier for your eyes. :wink:

alexlamsl:
@silverwind thanks for taking a look. I've expanded the tests using your example.

PR #812, opened 2017-05-22
Don't collapse whitespace inside elements that have white-space:pre as an inline style

papandreou:


alexlamsl:
I'm trying to understand the use case for this feature.

I'd expect `class` being the more common form of specifying this sort of whitespace rule (which we cannot detect, especially if the definition is on a separate style sheet).

Does this PR solve an issue which `conservativeCollapse` isn't able to cover?

papandreou:
I agree that `white-space: pre` hiding in an external stylesheet is more common than it showing up in an inline style. However, I believe that this is still a valid fix for html-minifier.

My use case is actually a `white-space: pre-wrap` in an external stylesheet, but it seems like it'll be easier to rewrite that to use an inline style than to expose the ability to specify `options.canTrimWhitespace` all the way through my stack :)

alexlamsl:
Wouldn't inlining `white-space: pre-wrap` make the size of your HTML file increase significantly?

Would that offset the byte savings when compared with `conservativeCollapse`?

I'm not too thrilled about inserting a very (CSS-)specific rule to `html-minifier`, especially if this isn't a common coding pattern to begin with.

If you don't want extra code complexity, you may consider `conservativeCollapse:true` globally and forgo saving a few bytes.

Having said that, I'm wondering if extending `collapseWhitespace` to be more configurable would be the best of both worlds. That way, you can even specify a match with your particular CSS class name without re-coding your HTML files.

papandreou:
> Wouldn't inlining white-space: pre-wrap make the size of your HTML file increase significantly?

In my case, no. That inline style would sit in a node quite far up the tree.

> Would that offset the byte savings when compared with conservativeCollapse?

`conservativeCollapse` is not enough in my case (I'm actually already specifying that). My problem is that I'm minifying HTML where in some subtrees, multiple spaces are used for alignment of the contents. Even with `conservativeCollapse` turned on, multiple spaces are compressed into one, so that won't be sufficient.

> I'm not too thrilled about inserting a very (CSS-)specific rule to html-minifier, especially if this isn't a common coding pattern to begin with.

It might not be a common coding pattern (I don't know), but I think it falls into a category where html-minifier obviously does something wrong, and it has all the information to detect it.

> Having said that, I'm wondering if extending collapseWhitespace to be more configurable would be the best of both worlds. That way, you can even specify a match with your particular CSS class name without re-coding your HTML files.

I thought about doing something like that (actually, to alter the behavior for elements matching arbitrary CSS selectors), but it looked like it would be very hard to adapt the html-minifier code to do that. A more limited feature triggered by a class being present would also solve my original problem, so I'd be in favor of that :)

papandreou:
Hmm, it actually occurred to me that specifying `options.canTrimWhitespace` and `options.canCollapseWhitespace` as a function that returns false won't be sufficient due to the "or" constructs here: https://github.com/kangax/html-minifier/blob/19845f3d141bea1eb2adf693b06b6bf6d6e9c4ff/src/htmlminifier.js#L897-L901

... If the built-in implementation of `can{Trim,Collapse}Whitespace` returns true, the user-specified one is never consulted, so it's not possible to veto the built-in one returning true.

Since `options.can{Trim,Collapse}Whitespace` are already [being defaulted](https://github.com/kangax/html-minifier/blob/gh-pages/src/htmlminifier.js#L629-L636), maybe it would be enough to check `options.can{Trim,Collapse}Whitespace()` in lines 897 and 901? And perhaps the default one could be provided as the third parameter so that users wouldn't need to reimplement the `<pre>`, `<textarea>` logic when overriding?

papandreou:
> Since `options.can{Trim,Collapse}Whitespace` are already being defaulted

Oh, it wasn't actually being defaulted, so I would suggest changing it like this: https://github.com/papandreou/html-minifier/commit/67d608fece5ffa0bd826effd04efa7fb335b4da7

alexlamsl:
I see, so you need a behaviour like `<pre>` as opposed to `conservativeCollapse`.

Have you considered `<!-- htmlmin:ignore -->`?

While the code change looks minimal by itself, I'm more concerned with cost of maintenance going forward. Since this is completely non-extensible and is expected to be rarely used, I'd be in favour of existing solutions if at all possible.

The bigger question is obviously whether `html-minifier` should take all CSS rules into account when performing optimisations, and if so how should we approach this in a manageable manner.

alexlamsl:
> Oh, it wasn't actually being defaulted, so I would suggest changing it like this: papandreou/html-minifier@67d608f

If that patch alone would address you needs, then I am not against it since it does not add much overhead.

papandreou:
> If that patch alone would address you needs, then I am not against it since it does not add much overhead.

Okay, you've got yourself a deal. Opened it as a separate PR here: https://github.com/kangax/html-minifier/pull/813

For the record I still think this PR is a relevant correctness fix to pull in, but you're the boss :)

alexlamsl:
Thanks for the understanding :wink: 

I'll keep this PR open to keep track on this potential area of improvement.

papandreou:
Okay, that's cool! I've rebased it and fixed the conflict it had with the other change.

PR #813, opened 2017-05-22
Add the ability for options.can{Trim,Collapse}Whitespace to veto the default implementation when that returns true.

papandreou:
Also, provide the default implementation as the 3rd parameter to the custom hook so the user doesn't have to reimplement it.

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -626,13 +626,11 @@ function processOptions(options) {
     options.log = identity;
   }
 
-  var defaultTesters = ['canCollapseWhitespace', 'canTrimWhitespace'];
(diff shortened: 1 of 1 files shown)


papandreou:
Sorry to be a PITA, but approx. when can I expect an npm release with this change? Otherwise I'll just publish a fork.

alexlamsl:
I'm building `3.5.1` as we speak, so Soon:tm:

alexlamsl:
`html-minifier 3.5.1` published on `npm`

papandreou:
Thanks a bunch. Really appreciate the quick assistance wrt. all of this :)

PR #817, opened 2017-05-24
Fix stackNoTrimWhitespace and stackNoCollapseWhitespace housekeeping

papandreou:
Previously the first encountered end tag matching the top item would clear it from the stack, regardless of whether the two were actually a start + end pair.

Consider the document:

```html
<div id=foo><div></div>foo  bar</div>
```

If `options.canTrimWhitespace` returned false for `<div id=foo>`, that would push `'div'` onto `stackNoTrimWhitespace`, but it would be cleared again when the first `</div>` is encountered, and thus `foo  bar` would have its whitespace squeezed so it comes out as `foo bar`. Since it works fine without the `<div></div>`.

My suggested fix is to add every encountered tag to `stackNoTrimWhitespace` (and `stackNoCollapseWhitespace`) when the stack isn't already empty. That makes sure that the end tags always pair up correctly with the stack items.

### DIFF
--- a/tests/minifier.js
+++ b/tests/minifier.js
@@ -3245,3 +3245,43 @@ QUnit.test('tests from PHPTAL', function(assert) {
     }), tokens[0]);
   });
 });
+
+QUnit.test('canCollapseWhitespace and canTrimWhitespace hooks', function(assert) {
+  function canCollapseAndTrimWhitespace(tagName, attrs, defaultFn) {
+    if ((attrs || []).some(function(attr) { return attr.name === 'class' && attr.value === 'leaveAlone'; })) {
+      return false;
+    }
+    return defaultFn(tagName, attrs);
+  }
+
+  var input = '<div class="leaveAlone"><span> </span> foo  bar</div>';
+  var output = '<div class="leaveAlone"><span> </span> foo  bar</div>';
+
+  assert.equal(minify(input, {
+    collapseWhitespace: true,
+    canTrimWhitespace: canCollapseAndTrimWhitespace,
+    canCollapseWhitespace: canCollapseAndTrimWhitespace
+  }), output);
+
+  // Regression test: Previously the first </div> would clear the internal
+  // stackNo{Collapse,Trim}Whitespace, so that ' foo  bar' turned into ' foo bar'
+  input = '<div class="leaveAlone"><div></div><span> </span> foo  bar</div>';
+  output = '<div class="leaveAlone"><div></div><span> </span> foo  bar</div>';
+
+  assert.equal(minify(input, {
+    collapseWhitespace: true,
+    canTrimWhitespace: canCollapseAndTrimWhitespace,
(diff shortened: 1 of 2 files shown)


tests/minifier.js
```diff
@@ -3245,3 +3245,29 @@ QUnit.test('tests from PHPTAL', function(assert) {
...
+    canTrimWhitespace: canTrimWhitespace
+  }), output);
+
+  // Regression test: Previously the first </div> would clear the internal stackNoTrimWhitespace,
+  // so that 'foo  bar' would be turned into 'foo bar'
+  input = '<div class="doNotTrimWhitespace"><div></div> foo bar </div>';
+  output = '<div class="doNotTrimWhitespace"><div></div> foo bar </div>';
+
+  assert.equal(minify(input, {
+    trimWhitespace: true,
```
alexlamsl: Do you mean `collapseWhitespace` here?

tests/minifier.js
```diff
@@ -3245,3 +3245,29 @@ QUnit.test('tests from PHPTAL', function(assert) {
...
+  }
+  assert.equal(minify(input, {
+    trimWhitespace: true,
+    canTrimWhitespace: canTrimWhitespace
+  }), output);
+
+  // Regression test: Previously the first </div> would clear the internal stackNoTrimWhitespace,
+  // so that 'foo  bar' would be turned into 'foo bar'
+  input = '<div class="doNotTrimWhitespace"><div></div> foo bar </div>';
+  output = '<div class="doNotTrimWhitespace"><div></div> foo bar </div>';
```
alexlamsl: ~Shouldn't the whitespaces in `</div> foo bar </div>` be optimised here?~

Edit: Never mind - I misread `</div>` as `<div>`. But please add a positive test case:
```html
<div class="doNotTrimWhitespace"></div><div> foo bar </div>
````

tests/minifier.js
papandreou: To be honest I had some trouble figuring out exactly what the difference between trimWhitespace and collapseWhitespace are. In my own app where I ran into this problem I found out I always want to toggle them on and off together.

alexlamsl: `trimWhitespace` is not a valid option - it's just a private function within the codebase.

alexlamsl: `html-minifier` does not throw error upon invalid options - it just ignores them.

papandreou: Oh, you're right. I was probably confused because `canTrimWhitespace` is accepted as an option. So what I actually found out was that in my app I had to specify `canTrimWhitespace` and `canCollapseWhitespace` together :)

tests/minifier.js
papandreou: Sure thing. Added the positive test case, squashed and force-pushed now.

alexlamsl:
Looks good, thanks!

I'll make a new release later today when I get home.

papandreou:
Thanks a lot for your hard work :)

PR #828, opened 2017-06-26
Implement the --inplace flag.

shlomif:
With a rudimentary test that isn't invoked by default. This is similar to the -i flag of perl, ruby or GNU sed in that it modifies existing files.


alexlamsl:
Does specifying the same input & output file not working thus requiring this new flag?

shlomif:
On Mon, 26 Jun 2017 19:30:28 -0700
"Alex Lam S.L." <notifications@github.com> wrote:

> Does specifying the same input & output file not working thus requiring this
> new flag?
> 

Hi!

Well, I'd expect it to work, but it can only handle one file at a time, whereas
--inplace can handle multiple files similar to «perl -pi -e
's#str#replacement#g' *.html [more files here]». Needing to invoke an
executable separately for each and every file may be prohibitive speed-wise, for
instance for this use case:

https://github.com/shlomif/shlomi-fish-homepage/commit/3bb006ec2fc752e90891ad2c655345df84831eb0

Regards,

-- @shlomif

shlomif:
Ping! Any news? Is there anything else I can do?

alexlamsl:
This PR does not pass Travis CI: https://travis-ci.org/kangax/html-minifier/builds/247290811

On the code changes itself, there is no need to introduce `inplace` as `commander` already stores the `--inplace` flag. Also I fail to see why a simple setting of output file/directory to that of input wouldn't suffice, rather than all the extra file handling code.

shlomif:
On Sun, 16 Jul 2017 21:05:05 -0700
"Alex Lam S.L." <notifications@github.com> wrote:

> This PR does not pass Travis CI:
> https://travis-ci.org/kangax/html-minifier/builds/247290811
> 

I'll try to fix it.

> On the code changes itself, there is no need to introduce `inplace` as
> `commander` already stores the `--inplace` flag.

Can it be accessed the same way?

> Also I fail to see why a
> simple setting of output file/directory to that of input wouldn't suffice,
> rather than all the extra file handling code.
> 

Well, with an output file there is only one that can be specified and you may
wish to process several files in one invocation. An output directory is only
valid if there is an input directory, and in this case all files inside the
directory and its subdirectory will be processed, while I want to avoid
processing some of them or only process a few selected ones. It is also
sometimes convenient to specify "--inplace" from the command line.

Regards,

	@shlomif

shlomif:
Travis-CI fixed now.

Issue #843, opened 2017-08-27
It seems like html-minifier does not support es6?


drmrbrewer:
Probably because html-minifier uses [uglify-js](https://www.npmjs.com/package/uglify-js), and uglify-js doesn't support es6... but [uglify-es](https://www.npmjs.com/package/uglify-es) *does*... see [here](https://github.com/mishoo/UglifyJS2/issues/659).

Can we switch to using uglify-es in html-minifier too?  How?

drmrbrewer:
Anyone?

alexlamsl:
Use `minifyJS: function(text, inline) { ... }` to swap in any external JavaScript optimiser as you see fit.

`uglify-es` is pretty much in the alpha stage, without any extensive test coverage etc. so won't be used as the default minifier here.

drmrbrewer:
@alexlamsl what is the `inline` argument?  Do you have a simple example (even pseudo-example) to help me implement this?

alexlamsl:
`inline=true` is for things like `<a onclick="doIt()">` - normal `<script>alert(42);</script>` has `inline=false`

drmrbrewer:
@alexlamsl  sorry I need a bit more hand-holding.  At present I'm specifying a config file via the `-c` switch.  And in the config file I have a bunch of settings like the following:

    {
      "caseSensitive": false,
      "minifyJS": true,
      "minifyCSS": true
    }

So just to test this, I tried providing a function for `minifyJS` that just returns `text` unaltered:

    {
      "caseSensitive": false,
      "minifyJS": "function(text, inline) { return text }",
      "minifyCSS": true
    }

But when I run `html-minifier` with the above config, it seems like it is still using the default js minifier?

alexlamsl:
You'll need to use the API rather than CLI to specify `function` as parameter values.

drmrbrewer:
Ah, therein lies the problem.  I need to use it via the CLI.  Is it not possible to add a simple option to `minifyJS` to allow you to specify es6 minification so that it's usable via the CLI?  

And is `uglify-es` really basically alpha?  I note that you're the author / publisher on that one too.  And I note that there are 170,000 downloads of `uglify-es` every month, without any mention or warning in the README that it's only alpha quality...

alexlamsl:
I don't actively support `uglify-es` - it is there to make it easier for other people to test and develop, which hasn't really happened much so far.

Feel free to make a pull request to extend CLI support for `function` type parameter values.

[...]

PR #846, opened 2017-09-01
added ES6 minification support

Ja9ad335h:
to support `ES6` minification inside `<script>` blocks

- replaced `uglify-js` with `uglify-es`
- replaced `grunt-contrib-uglify` with `grunt-contrib-uglify#harmony` (for testing)


alexlamsl:
See https://github.com/kangax/html-minifier/issues/843#issuecomment-326791250

cekvenich:
So... it should now work with es6?

PR #849, opened 2017-09-11
Preserve en/em/non-breaking/hair space etc. while minifying

papandreou:
There are some "precious" space characters matched by /\s/ that should never be collapsed or trimmed away.

See https://github.com/assetgraph/assetgraph/pull/778

### DIFF
--- a/tests/minifier.js
+++ b/tests/minifier.js
@@ -370,6 +370,10 @@ QUnit.test('types of whitespace that should always be preserved', function(asser
 
   // Non-breaking space passed as HTML entity, in decodeEntities:true mode:
   assert.equal(minify(inputWithEntities, { collapseWhitespace: true, decodeEntities: true }), input);
+
+  // Do not remove hair space when preserving line breaks between tags:
+  input = '<p></p>\u200a\n<p></p>\n';
+  assert.equal(minify(input, { collapseWhitespace: true, preserveLineBreaks: true }), input);
 });
 
 QUnit.test('doctype normalization', function(assert) {
--- a/tests/minifier.js
+++ b/tests/minifier.js
@@ -347,6 +347,31 @@ QUnit.test('space normalization around text', function(assert) {
   assert.equal(minify(input, { collapseWhitespace: true }), output);
 });
 
+QUnit.test('types of whitespace that should always be preserved', function(assert) {
+  // Hair space:
+  var input = '<div>\u200afo\u200ao\u200a</div>';
+  assert.equal(minify(input, { collapseWhitespace: true }), input);
+
+  // Hair space passed as HTML entity:
+  var inputWithEntities = '<div>&#8202;fo&#8202;o&#8202;</div>';
+  assert.equal(minify(inputWithEntities, { collapseWhitespace: true }), inputWithEntities);
+
+  // Hair space passed as HTML entity, in decodeEntities:true mode:
+  assert.equal(minify(inputWithEntities, { collapseWhitespace: true, decodeEntities: true }), input);
+
+
+  // Non-breaking space:
+  input = '<div>\xa0fo\xa0o\xa0</div>';
+  assert.equal(minify(input, { collapseWhitespace: true }), input);
+
+  // Non-breaking space passed as HTML entity:
+  inputWithEntities = '<div>&nbsp;fo&nbsp;o&nbsp;</div>';
+  assert.equal(minify(inputWithEntities, { collapseWhitespace: true }), inputWithEntities);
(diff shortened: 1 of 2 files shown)


papandreou:
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#special-white-space `\s` is equivalent to `[ \f\n\r\t\v\u00a0\u1680\u180e\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]`.

Maybe the `[ \n\r\t]` character class I suggest here should be tweaked further to include more characters.

[The White Space Processing Rules](https://drafts.csswg.org/css-text-3/#white-space-processing) says:

> Control characters (Unicode category Cc) other than tab (U+0009), line feed (U+000A), form feed (U+000C), and carriage return (U+000D) must be rendered as a visible glyph and otherwise treated as any other character of the Other Symbols (So) general category and Common script. The UA may use a glyph provided by a font specifically for the control character, substitute the glyphs provided for the corresponding symbol in the Control Pictures block, generate a visual representation of its codepoint value, or use some other method to provide an appropriate visible glyph. As required by [UNICODE], unsupported Default_ignorable characters must be ignored for rendering.

... so at least form feed (`\f`) should also be added.

alexlamsl:
I prefer only changing the `/\s/` that is necessary to pass your new unit tests rather than a blanket replacement.

For instance, https://github.com/kangax/html-minifier/pull/849/files#diff-bd35077e6c438d3f6866b57bb2481260L33 isn't necessary.


(I have tested on a few web browsers and agree that the issue illustrated by your test cases need to be fixed.)

papandreou:
> I prefer only changing the /\s/ that is necessary to pass your new unit tests rather than a blanket replacement.
> For instance, https://github.com/kangax/html-minifier/pull/849/files#diff-bd35077e6c438d3f6866b57bb2481260L33 isn't necessary.

Just to be sure that I understand your POV correctly -- isn't that actually an argument for adding test cases that validate that and the other replacement?

alexlamsl:
@papandreou I see - if you can have the new test cases cover all the replacements of `\s` to `[ \r\n\t\f]`, I'll be very happy and convinced :wink:

papandreou:
Okay, I added some more tests for the cases I could find, then rolled back to `\s` in the remaining cases.

However, I'm fairly sure that some of those tests can be written, so I probably wouldn't include https://github.com/kangax/html-minifier/pull/849/commits/1d600ccdeb0d47ec8ad737997e467eb1589401ed -- those other kinds of spaces should basically be treated as regular glyphs, so I think `/[ \n\r\t\f]/` is a safer default.

alexlamsl:
I prefer changing them only if/when we hit an issue, and would therefore has test coverage for it.

The new tests & rollbacks LGTM - if there isn't anything else I shall go ahead & merge.

tests/minifier.js
```diff
@@ -347,6 +347,44 @@ QUnit.test('space normalization around text', function(assert) {
...
+  // Non-breaking space passed as HTML entity, in decodeEntities:true mode:
+  assert.equal(minify(inputWithEntities, { collapseWhitespace: true, decodeEntities: true }), input);
+
+  // Do not remove hair space when preserving line breaks between tags:
+  input = '<p></p>\u200a\n<p></p>\n';
+  assert.equal(minify(input, { collapseWhitespace: true, preserveLineBreaks: true }), input);
+
+  // Preserve hair space in attributes:
+  input = '<p class="foo\u200abar"></p>';
+  assert.equal(minify(input, { collapseWhitespace: true, preserveLineBreaks: true }), input);
```
alexlamsl: `preserveLineBreaks` doesn't seem to be required here?

papandreou: Nice catch, fixed + rebased.

papandreou:
> if there isn't anything else I shall go ahead & merge

Nothing further for now. I might pick it up again when I get some time :)

alexlamsl:
@papandreou I shall merge after your rebase is pushed out then (GitHub having a slow day?)

Thanks a lot for the patch!

papandreou:
@alexlamsl Hah, I don't know where I tried to push to before, but here it is :)

papandreou:
@alexlamsl Thanks! Can I persuade you to do a new release with this fix?

alexlamsl:
That's the plan - I went straight to bed after merging your PR :sweat_smile: 

Give me a few hours while I sort out my $day_job, then I'll make `3.5.5` :wink:

PR #851, opened 2017-10-04
Remove \t in customFragments tokens

Vasikaran:
I am using **html-loader** and I found an issue on that. [Issue](https://github.com/webpack-contrib/html-loader/issues/148)

While I researching about that I just saw your codes. You have replaced customFragments with equals tokens of that in **htmlminifier.js**. But you have added tabs(\t) before and after the tokens. I have just removed these tabs(\t). After removing my issue fixed. I don't know If my changes are worth. But you must consider in these cases. Thank you!

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -880,6 +880,8 @@ function minify(value, options, partialMarkup) {
       }
       var token = uidAttr + ignoredCustomMarkupChunks.length;
       ignoredCustomMarkupChunks.push(/^(\s*)[\s\S]*?(\s*)$/.exec(match));
+      // return '\t' + token + '\t';
+      // remove \t in token
       return token;
     });
   }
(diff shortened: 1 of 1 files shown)


alexlamsl:
Please address the failed test cases - they illustrate why your proposed code change is unsafe.

Vasikaran:
Yeah, I noticed that. This is also use case. This small space has broken my webapp lots of places. What can I do?

alexlamsl:
`${module}-count` is by no means valid HTML attribute, so you may need to look at `ignoreCustomFragments` to include the whole token not just the `${...}` bit.

Vasikaran:
Okay I understood. Hereafter I will use ingoreCustomFragments properly.

alexlamsl:
I assume the issue has been addressed, so I'll close this PR.

Issue #798, opened 2017-04-06
Despite minifyCss set to false, inlined styles gets modified


PEM--:
When minifying this with `minifyCss: false`:
```
<p style="background-color:blue;color:red;">Test</p>
```
The output is changed and the last `;` in the inline style gets removed leading to reconciliation errors in React. While removing the last `;` is a neat feature, when the `minifyCss` is `false`, the inline style should not be altered.

Referenced errors:

* https://github.com/facebook/react/issues/9353
* https://github.com/ssrwpo/ssr/issues/45

alexlamsl:
Sounds reasonable - PR is welcomed.

petermikitsh:
This creates integration issues with react server-sider rendering. Right now, I'm only observing those issues in Firefox. The markup is reusable in Chrome + Safari.

```
Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:
 (client) ="background:#0C8DCE;" data-reactid="5">
 (server) ="background:#0C8DCE" data-reactid="5"><
```

I guess those trailing semicolons do matter.

buildbreakdo:
@petermikitsh Have this issue in Chrome as well. Unclear how to configure html-minifier to leave trailing semicolon alone in the style attribute, @alexlamsl is the referenced issue a proposed fix?

DChristiany:
I have created a fork (found here: https://github.com/DChristiany/html-minifier), that offers a configuration option, that does not remove trailing semicolons.
Should I just create a Pull Request (sorry, i'm new to working on open source projects).

PR #854, opened 2017-10-05
dont throw out trailing semicolon in style attribute

DChristiany:
See https://github.com/kangax/html-minifier/issues/798

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -283,7 +283,7 @@ function cleanAttributeValue(tag, attrName, attrValue, options, attrs) {
     attrValue = trimWhitespace(attrValue);
     if (attrValue) {
       if (/;$/.test(attrValue) && !/&#?[0-9a-zA-Z]+;$/.test(attrValue)) {
-        attrValue = attrValue.replace(/\s*;$/, '');
+        attrValue = attrValue.replace(/\s*;$/, ';');
       }
       attrValue = unwrapInlineCSS(options.minifyCSS(wrapInlineCSS(attrValue)));
     }
--- a/tests/minifier.js
+++ b/tests/minifier.js
@@ -877,11 +877,11 @@ QUnit.test('cleaning class/style attributes', function(assert) {
   assert.equal(minify(input), output);
 
   input = '<p style="    color: red; background-color: rgb(100, 75, 200);  "></p>';
-  output = '<p style="color: red; background-color: rgb(100, 75, 200)"></p>';
+  output = '<p style="color: red; background-color: rgb(100, 75, 200);"></p>';
   assert.equal(minify(input), output);
 
   input = '<p style="font-weight: bold  ; "></p>';
-  output = '<p style="font-weight: bold"></p>';
+  output = '<p style="font-weight: bold;"></p>';
   assert.equal(minify(input), output);
 });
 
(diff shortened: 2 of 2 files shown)


alexlamsl:
It would make more sense to suppress the trailing semicolon trimming when `minifyCSS:false` than to create a new flag that adds semicolons.

DChristiany:
We do want the minifying of the CSS, removing whitespace etc.
We just need the trailing semicolon.

I adapted the code, to make sure the semicolon is only added, when minifyCSS is true and adapted the tests.

alexlamsl:
So the only change for `html-minifier` I can see is when `minifyCSS: false`, currently it will still strip out the trailing semicolon - and I'm amenable to change that behaviour so that any style strings stays unmodified (the quotes may still be removed, but that's a different issue).

For the case when `minifyCSS` is enabled, you will need to find out the configuration for `clean-css` to preserve the trailing semicolon.

DChristiany:
You're right, I'll close this request

DChristiany:
I have gone through the code again and am convinced, that the html minifier removes the semicolon, before clean-css has a chance to.

Within cleanAttributeValue there are the following lines:
`if (/;$/.test(attrValue) && !/&#?[0-9a-zA-Z]+;$/.test(attrValue)) {
        attrValue = attrValue.replace(/\s*;$/, '');
}`
and the semicolon is removed within the call to replace.

alexlamsl:
Refer to the first paragraph of https://github.com/kangax/html-minifier/pull/854#issuecomment-334737124 for suggested fix to be included in this PR.

DChristiany:
Would this be a suitable implementation?

alexlamsl:
AFAICT your latest patch just adds trailing semicolon to all style attributes regardless of settings, so no that since it increases the size of minified output in all cases.

DChristiany:
No, I force pushed a new commit, replacing the old one

src/htmlminifier.js
```diff
@@ -283,7 +283,7 @@ function cleanAttributeValue(tag, attrName, attrValue, options, attrs) {
     attrValue = trimWhitespace(attrValue);
     if (attrValue) {
       if (/;$/.test(attrValue) && !/&#?[0-9a-zA-Z]+;$/.test(attrValue)) {
-        attrValue = attrValue.replace(/\s*;$/, '');
+        attrValue = attrValue.replace(/\s*;$/, ';');
```
alexlamsl, line 286: This will preserve the trailing semi-colon regardless of `minifyCSS`, which is what I meant in https://github.com/kangax/html-minifier/pull/854#issuecomment-338923832

The trailing semi-colon is not needed for standard HTML, so if you need to preserve it, only do so for `minifyCSS:false`

alexlamsl:
Please exclude `dist/*` from your commits - they will be re-generated during the next release.

src/htmlminifier.js
DChristiany: THis happens before minifyCss is executed, so minifyCss will throw out semicolons (unless configured otherwise) if minifyCss is true.
Should I add the check for minifyCSS === false anyways, to make it more explicit?

alexlamsl: I see what you mean - yes that is acceptable, as we'll now only trim whitespaces before trailing semi-colon, and `clean-css` takes care of the rest.

To make sure this works in the future, can you add a trailing semi-colon to the following test?
https://github.com/kangax/html-minifier/blob/9a45eb435a8a3ec453da52a752c0d4cf5c071b1d/tests/minifier.js#L2225

alexlamsl:
@DChristiany thanks for working on this - please address https://github.com/kangax/html-minifier/pull/854#discussion_r146529311 & https://github.com/kangax/html-minifier/pull/854#issuecomment-338947775 and we should be good to go.

DChristiany:
I Just force pushed a new version

hinok:
@kangax @alexlamsl Could you publish new version with this change to npm? :)

Issue #888, opened 2018-02-28
'ul' should belongs to non-Phrasing Content


weixu365:
ul is missing in non-Phrasing Content in https://github.com/kangax/html-minifier/blob/0f2a67faf4b347bfd6f8bf56ff43847547882ebb/src/htmlparser.js#L83

Here's the testing html
```
const html = `
<p>
  before ul
<ul>
  <li>item1</li>
  <li>item2</li>
</ul>
</p>`;
console.log(minify(html, { collapseWhitespace: true, conservativeCollapse: true }));
```

**Should return**
```
 <p> before ul </p><ul> <li>item1</li> <li>item2</li> </ul> <p></p>
```
This is also the default browser behaviour, and python beautifulSoup has the same logic

**But currently get**
```
<p> before ul <ul> <li>item1</li> <li>item2</li> </ul> </p>
```

I'm happy to create a pull request if you think it's a valid issue.

alexlamsl:
The current output renders exactly the same on the browser:
```html
<html>
<body>
<table border=1>
<tr>
<th>Input</th>
<th>Output</th>
</tr>
<tr>
<td>

<p>
  before ul
<ul>
  <li>item1</li>
  <li>item2</li>
</ul>
</p>

</td>
<td>

<p> before ul <ul> <li>item1</li> <li>item2</li> </ul> </p>

</td>
</tr>
</table>
```
![render](https://user-images.githubusercontent.com/1898413/36774279-44f1899e-1c55-11e8-8a29-d086d6fe9d08.png)

weixu365:
It would be easier to check the browser result if you **add some styles** or **using the inspector** to check the element tree. Here's example code:
```
<p style="margin: 50px;">
  testing ul in p
<ul>
  <li>item</li>
</ul>
after ul
</p>
```
It has the same behaviour on Chrome, Firefox and Safari under MacOSX, here's the screen shot and element tree
<img width="337" alt="ul-inside-p" src="https://user-images.githubusercontent.com/20695333/36778416-bf001820-1cc0-11e8-9eeb-508a7b3cf2dd.png">

The style only applied to the first part of **p** tag.

DamonHD:
The W3C validator does not like &lt;ul>&lt;/ul> inside &lt;p> ... &lt;/p> and it is regarded as incorrect, so should likely behave as @weixu365 expects, IMHO.

Rgds

Damon

alexlamsl:
[...]

PR #889, opened 2018-03-02
Add "ol" and "ul" to nonPhrasing

weixu365:
Closing issue #888.

I didn't change any tests in this PR because there's no current test covering which tag belongs to non-phrasing tags.

### DIFF
--- a/tests/minifier.js
+++ b/tests/minifier.js
@@ -1690,6 +1690,31 @@ QUnit.test('phrasing content', function(assert) {
   assert.equal(minify(input, { html5: true }), input);
 });
 
+// https://github.com/kangax/html-minifier/issues/888
+QUnit.test('ul/ol should be phrasing content', function(assert) {
+  var input, output;
+
+  input = '<p>a<ul><li>item</li></ul>';
+  output = '<p>a</p><ul><li>item</li></ul>';
+  assert.equal(minify(input, { html5: true }), output);
+
+  output = '<p>a<ul><li>item</ul>';
+  assert.equal(minify(input, { html5: true, removeOptionalTags: true }), output);
+
+  output = '<p>a<ul><li>item</li></ul></p>';
+  assert.equal(minify(input, { html5: false }), output);
+
+  input = '<p>a<ol><li>item</li></ol></p>';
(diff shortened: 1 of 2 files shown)


alexlamsl:
That's not how testing works.

Please add tests which fails in the current version but passes with your changes in this PR.

alexlamsl:
Please exclude any changes to `dist/*` and `package*.json` from this PR.

tests/index.html
```diff
@@ -3,12 +3,12 @@
   <head>
     <meta charset="utf-8">
     <title>HTML Minifier Tests</title>
-    <link rel="stylesheet" href="https://code.jquery.com/qunit/qunit-2.5.0.css">
+    <link rel="stylesheet" href="https://code.jquery.com/qunit/qunit-2.5.1.css">
```
alexlamsl: Please also exclude changes to this file.

tests/minifier.js
```diff
@@ -1690,6 +1690,21 @@ QUnit.test('phrasing content', function(assert) {
...
+
+  input = '<p>a<ul><li>item</li></ul>';
+  output = '<p>a</p><ul><li>item</li></ul>';
+  assert.equal(minify(input, { html5: true }), output);
+  output = '<p>a<ul><li>item</li></ul></p>';
+  assert.equal(minify(input, { html5: false }), output);
+
+  input = '<p>a<ol><li>item</li></ol></p>';
+  output = '<p>a</p><ol><li>item</li></ol><p></p>';
+  assert.equal(minify(input, { html5: true }), output);
```
alexlamsl, line 1709: Tests look good, thanks.

Can you add one more that also does `removeOptionalTags: true` to make sure the newly placed `</p>` doesn't get removed?

weixu365:
Sorry for added dist by mistake.

tests/minifier.js
weixu365: The generated &lt;/p> before ul/ol should be removed because it's optional. I also added test to check the &lt;p>&lt;/p> at the end.

alexlamsl: Perfect - thanks! :+1:

Issue #904, opened 2018-04-07
Whitespace around <ruby> is removed


YodasWs:
`<ruby>` is an inline tag, yet the minifier is removing all whitespace that surrounds it.

### Input
```html
<p>I love to eat <ruby lang="fr">pain au chocolat<rt lang="en-US">chocolate croissants</rt></ruby> and éclairs.</p>
```

### Expected
I would not expect the above line to have any changes.

### Actual
I am actually getting this output:
```html
<p>I love to eat<ruby lang="fr">pain au chocolat<rt lang="en-US">chocolate croissants</rt></ruby> and éclairs.</p>
```
Which would get displayed like this in a browser:
```txt
I love to eatpain au chocolatand éclairs.
```

## `<rp>`
:warning: You will also need to test and potentially correct whitespace in/around `<rp>`. E.g.:
```html
<p>I love to eat <ruby lang="fr">pain au chocolat<rp> (</rp><rt lang="en-US">chocolate croissants</rt><rp>)</rp></ruby> and éclairs.</p>
```
ℹ️ Notice the space before the opening parenthesis in the first `<rp>`.

## also
This was found in `html-minifier` version `3.5.6`.

alexlamsl:
Pull Request welcomed.

PR #905, opened 2018-04-07
Preserve whitespace around inline <ruby>

YodasWs:
also:
* Preserve whitespace inside `<rp>` tag (to buffer parentheses for very old browsers)
* Preserve whitespace around inline `<rtc>` tags
* Mark `<rb>` as valid HTML

for #904

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -67,9 +67,9 @@ function collapseWhitespace(str, options, trimLeft, trimRight, collapseAll) {
 
 var createMapFromString = utils.createMapFromString;
 // non-empty tags that will maintain whitespace around them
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -7,3 +7,5 @@
 /benchmarks/generated
 /node_modules/
 /npm-debug.log
+*.swp
+*~
```
alexlamsl: Please leave this out of the PR.

alexlamsl (changes requested):
Code and test changes are good to go.

alexlamsl (approved):

.gitignore
alexlamsl: Never mind - I have performed the edit.

alexlamsl:
Thanks for the work!

PR #907, opened 2018-04-11
Callback Support

### LINKED ISSUE #906: Add callback to minifyCSS and minifyJS functions to allow for asynchronous calls.
Currently the functions `minifyCSS` and `minifyJS` cannot perform any asynchronous calls. It would be a really nice enhancement to add callback support.

### Example of possible new syntax

```js
var minify = require('html-minifier').minify;
var html = '<html><head><style>body { margin: 0; display: flex; }</style></head></html>';

var result = minify(html, {
  minifyCSS: function(css, callback) {
    doSomethingAsync(css, function(updatedCss) {
      callback(updatedCss);
    });
  }
});
```

This enhancement can be added without any changes to the way synchronous code is executed.

Related: #900

RebeccaStevens:
Adds callback support to `minify`, `minifyCSS` and `minifyJS` functions.
Does not break any existing functionality.

### Example Use:

```js
var minify = require('html-minifier').minify;
var html = '<html><head><style>body { margin: 0; display: flex; }</style></head></html>';

var config = {
  minifyCSS: function(css, cb) {
    doSomethingAsync(css, function(updatedCss) {
      cb(updatedCss);
    });
  },
  minifyJS: function(js, inline, cb) {
    doSomethingElseAsync(js, function(updatedJs) {
      cb(updatedJs);
    });
  }
};

minify(html, config, function(error, result) {
  if (error) {
    // Handle error.
  } else {
    // Do something with the results.
  }
});
```

Fixes #906

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -1176,7 +1176,7 @@ function minify(value, options, partialMarkup, cb) {
           tasksWaitingFor++;
           var minifyJSResult = options.minifyJS(text, null, onTaskFinished);
 
-          // If the result is defined then minifyJSResult completed synchronously.
+          // If the result is defined then minifyJS completed synchronously.
           // eslint-disable-next-line no-undefined
           if (minifyJSResult !== undefined) {
             onTaskFinished(minifyJSResult);
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -941,6 +941,7 @@ function minify(value, options, partialMarkup, cb) {
     trimTrailingWhitespace(charsIndex, nextTag);
   }
 
+  var str;
   new HTMLParser(value, {
     partialMarkup: partialMarkup,
     html5: options.html5,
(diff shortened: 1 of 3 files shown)


RebeccaStevens:
I plan on adding documentation updates soon.

alexlamsl:
As stated in https://github.com/kangax/html-minifier/issues/906#issuecomment-379634151, you need to preserve the signature of `minify()`, i.e. it needs to return a string that is the minified input.

Any changes to that interface would be a breaking change and cannot be accepted.

RebeccaStevens:
The signature of `minify()` is preserve for synchronous calls. It only differs for asynchronous calls.

alexlamsl:
I see.

Before I can perform a code review, please exclude any auto-generated files from the changeset, i.e. `dist/*.js`

RebeccaStevens:
Changes to dist files have now been undone.

RebeccaStevens:
Just did a rebase.

alexlamsl:
Thanks - looking at the changes, one question jumps out - is `HTMLParser` still functional synchronously?

Apologies for not mentioning this earlier - there are users would rely on that to perform pre/post-processing, outside the scope of minification.

One way I can think of is similar to `ignoreCustomFragments`, i.e. assign temporary (unique) strings as placeholders for asynchronous `minify{CS,J}S`. That way the changes would also be confined within `src/htmlminifier.js`

src/htmlminifier.js
```diff
@@ -1160,42 +1161,85 @@ function minify(value, options, partialMarkup) {
...
-        // <body> may be omitted if first thing inside is not space, comment, <meta>, <link>, <script>, <style> or <template>
-        if (optionalStartTag === 'html' || optionalStartTag === 'body' && !/^\s/.test(text)) {
-          removeStartTag();
+        try {
+          tasksWaitingFor++;
+          var minifyCSSResult = options.minifyCSS(text, onTaskFinished);
+
+          // If the result is defined then minifyCSS completed synchronously.
+          // eslint-disable-next-line no-undefined
+          if (minifyCSSResult !== undefined) {
```
alexlamsl: Please use `typeof minifyCSSResult !== 'undefined'` instead of overriding `eslint`

RebeccaStevens: Will do.

RebeccaStevens:
`HTMLParser` is synchronous when `minifyCSS` and `minifyJS` are synchronous.

Does `HTMLParser` also need to be made synchronous for when either `minifyCSS` or `minifyJS` are asynchronous?

alexlamsl:
> `HTMLParser` is synchronous when `minifyCSS` and `minifyJS` are synchronous

I seem to have trouble validating that - the following test file is used:
```js
// test.js
var HTMLParser = require('./src/htmlparser').HTMLParser;
var input = require('fs').readFileSync('./index.html', 'utf8');

function log(prefix) {
  return function() {
    var args = [].slice.call(arguments);
    args.unshift(prefix);
    console.log.apply(console, args);
  };
}

new HTMLParser(input, {
  start: log('start'),
  end: log('end'),
  chars: log('chars'), 
  comment: log('comment')
});
```
On current HEAD of `gh-pages`:
```
$ cat test.js | node
chars
  html
start html [ { name: 'lang',
    value: 'en',
    customAssign: '=',
    customOpen: '',
    customClose: '',
    quote: '"' } ] false
chars
   html head
start head [] false
... snip many lines in between ...
end body [] false
chars
 /body /html
end html [ { name: 'lang',
    value: 'en',
    customAssign: '=',
    customOpen: '',
    customClose: '',
    quote: '"' } ] false
chars
 /html
```
... whereas with this PR:
```
$ cat test.js | node
chars
  html function (error) {
            if (error) {
              return handleError(error);
            }

            prevTag = '';
            checkForParseError();
          }
```

RebeccaStevens:
I'll look into that tomorrow and make the necessary change to get it working.

PR #923, opened 2018-05-22
chore: cache node_modules

DanielRuf:


alexlamsl:
What is the justification of this change, may I ask?

I see value in testing the latest versions of dependencies, in case any compatibility issues arises.

DanielRuf:
> I see value in testing the latest versions of dependencies, in case any compatibility issues arises.

Especially this as there are often issues with dependencies and Node.js 10.
Also current LTS and stable should be at least tested which are 8 and 10. The versions in between are not so important release branches.

DanielRuf:
It enables the cache to cache `node_modules` which produces faster builds.

Wait a second.

DanielRuf:
https://travis-ci.org/kangax/html-minifier/builds/382331691

https://travis-ci.org/kangax/html-minifier/builds/382388592

alexlamsl:
This PR shows similar performance as the current builds: https://travis-ci.org/kangax/html-minifier/builds/381384835

I am afraid the default settings also win this round.

DanielRuf:
Similar? 1:50 vs 2:20 minutes?

DanielRuf:
Also you did not benchmark with and without the cache, so how do you know that your build would not have been faster with the cache?

DanielRuf:
In sum, the complete build would be at least 1 minute faster (10:40 vs 9:44).

PR #924, opened 2018-05-22
chore: checkout last 5 commits

DanielRuf:


alexlamsl:
Thanks for the interesting in tweaking, but I think we'll stock to whatever default settings Travis CI has in this case.

DanielRuf:
> Thanks for the interesting in tweaking, but I think we'll stock to whatever default settings Travis CI has in this case.

By default Travis uses `50` which is often slower as there are more diffs to calculate than when you use `5`.

PR #929, opened 2018-05-26
Ignore custom fragments for minifyCSS

### LINKED ISSUE #928: minifyCSS and custom fragments
Hi, there is a bug using `minifyCSS` and `ignoreCustomFragments`, it seems that `clean-css` is processing fragments that should be ignored

Code:
```js
minify('<style>body{color: red;}{{ foo }}</style>', {
  minifyCSS: true,
  ignoreCustomFragments: [/{{.+}}/]
});
```

Expected output:
```html
<style>body{color:red}{{ foo }}</style>
```

Actual output:
```html
<style>body{color:red}</style>
```

For comparison, `minifyJS` works as expected:
```js
minify('<script>var a = true;{{ foo }}</script>', {
  minifyJS: true,
  ignoreCustomFragments: [/{{.+}}/]
});
// '<script>var a=!0;{{ foo }}</script>'
```

Using version 3.5.16

felipemsantana:
Closes #928

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -355,6 +355,15 @@ function unwrapMediaQuery(text) {
   return matches ? matches[1] : text;
 }
 
+// FIXME: Replace with https://github.com/jakubpawlowicz/clean-css#how-to-keep-a-css-fragment-intact when released
+function wrapIgnoredCSS(text) {
+  return '/*!htmlmin' + text + '*/';
+}
+
+function unwrapIgnoredCSS(text) {
+  return text.replace(/\/\*!htmlmin([\s\S]*?)\*\//, '$1');
+}
+
 function cleanConditionalComment(comment, options) {
   return options.processConditionalComments ? comment.replace(/^(\[if\s[^\]]+]>)([\s\S]*?)(<!\[endif])$/, function(match, prefix, text, suffix) {
     return prefix + minify(text, options, true) + suffix;
@@ -652,7 +661,7 @@ function processOptions(values) {
       }
       options.minifyCSS = function(text, type) {
         text = text.replace(/(url\s*\(\s*)("|'|)(.*?)\2(\s*\))/ig, function(match, prefix, quote, url, suffix) {
-          return prefix + quote + options.minifyURLs(url) + quote + suffix;
+          return prefix + quote + options.minifyURLs(unwrapIgnoredCSS(url)) + quote + suffix;
         });
         try {
           if (type === 'inline') {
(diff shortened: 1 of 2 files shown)


felipemsantana:
@alexlamsl as you can see here: https://github.com/kangax/html-minifier/pull/929/commits/4eb22fe53aa130e7c1afb6dbfce0bcc3266318e7
Not all tests passed using this implementation, I think the only way is changing the [clean-css tokenizer](https://github.com/jakubpawlowicz/clean-css/blob/master/lib/tokenizer/tokenize.js)

Should `specialComments: 'all'` be forced in `clean-css`?
Or maybe show a warning if the value is not `all`.

alexlamsl:
That will break as soon as there exists an instance of `specialComments` in the original code.

And if you try to use the numeral settings, the ordering of existing and injected `specialComments` would still break everything.

But as you stated in the code comment, upcoming `clean-css@4.2` will have `/* clean-css ignore */`, so may be we should re-evaluate the fix for that.

Note that the same custom fragment issue would apply to JavaScript snippets since `uglify-js` won't handle invalid syntax either - though it would throw an error which causes `html-minifier` to fall back to verbatim output.

May be this PR should just be picking up any tokenizer warnings from `clean-css` and output the verbatim styles instead of the incorrect minified output?

felipemsantana:
I agree about how to handle of warnings of `clean-css`, will do that.

I think we should show a warning about `specialComments` if the value is not `all`.

felipemsantana:
These tests are failing because they depends of `clean-css` tokenizer removal of invalid chars and it emits warnings :confused: 
https://github.com/kangax/html-minifier/blob/8043e97d4c961edf5dd938073643f94a56a180ed/tests/minifier.js#L645
https://github.com/kangax/html-minifier/blob/8043e97d4c961edf5dd938073643f94a56a180ed/tests/minifier.js#L693
https://github.com/kangax/html-minifier/blob/8043e97d4c961edf5dd938073643f94a56a180ed/tests/minifier.js#L3238

felipemsantana:
https://travis-ci.org/kangax/html-minifier/jobs/384392625

To avoid breaking backwards compatibility I suggest to add a new option called `ignoreWarnings` that defaults to `true`, in next major version of `html-minifier` you could switch to `false`, I think this is the only place that has warnings but it won't be strictly bound to `clean-css` and can be used everywhere else.

felipemsantana:
@alexlamsl could you please review the changes and my suggestion? I can't use the `minifyCSS` option because of this bug.

alexlamsl:
I couldn't see a way for this change to not break other existing use cases - may be `clean-css@4.2` will give us a viable option of dealing with custom fragments.

Until then, I'll leave this open until a proper solution comes to mind.

PR #930, opened 2018-05-29
Function uniqueId always returns a 12 chars id

alxscms:
The function uniqueId was making this library "impure". The output of the minifier could be different from one call to another using the same input and using the option `maxLineLength`.

Let me explain: 

The function `joinResultSegments` is relying on the segments length to decide whether to add a newline separator or not when joining them while using the option `maxLineLength`. As the function `uniqueId` was generating a random length ID, sometimes you could have different outputs with HTML elements being moved to a new line while it wasn't on a previous call.

I changed the code of `uniqueId` a bit so that the returned ID is always a 12 chars string.

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -726,7 +726,9 @@ function processOptions(values) {
 function uniqueId(value) {
   var id;
   do {
-    id = Math.random().toString(36).replace(/^0\.[0-9]*/, '');
+    id = 'xxxxxxxxxxxx'.replace(/x/g, function () {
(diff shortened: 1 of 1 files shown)


alexlamsl (changes requested):
Please provide a test case (which may not fail all the time due to the random nature of length returned by `uniqueId()`) before we proceed.

For instance, `minify('<div><?foo?></div>', { maxLineLength: 5 })` does not yield any difference with or without this change over a million runs.

PR #932, opened 2018-05-31
Do not collapse boolean attributes with ignored values.

### LINKED ISSUE #931: Values of boolean attributes should not be removed if they are ignored fragments.
I'm working on a babel plugin to help minify [lit-html](https://github.com/Polymer/lit-html) templates.  Minimized example of incorrect translation:
```js
import {html} from 'lit-html';

export default function(disabled) {
  return html`<input readonly="readonly" disabled="${disabled}">`;
}
```

The script is processed by a babel plugin which finds the `html` template, rewrites it to `<input readonly="readonly" disabled="${__BABEL HTML MINIFIER PLACEHOLDER__}">`.  This string is then processed with html-minifier.  I'm using the following html-minifier options:
```js
{
	collapseBooleanAttributes: true,
	collapseWhitespace: true,
	ignoreCustomFragments: [/\$\{__BABEL HTML MINIFIER PLACEHOLDER__\}/g],
}
```
This correctly removes the value of `readonly`, but (IMO) incorrect removes the value of `disabled`.  My goal is to have html-minifier collapseBooleanAttributes apply to static attributes only, since `lit-html` uses `disabled="${disabled}" to set/unset the property based on the value of the `disabled` argument.

### LINKED ISSUE #981: not working-Failed to construct 'Worker'
I am totally new to this...I opened the Index.html in the main directory--pasted my HTML code and clicked on the minify button
Nothing happens
In console it shows error:

master.js:21 Uncaught DOMException: Failed to construct 'Worker': Script at 'file:///C:/aPORTABLE%20Programs/html-minifier-gh-pages/assets/worker.js' cannot be accessed from origin 'null'.
    at file:///C:/aPORTABLE%20Programs/html-minifier-gh-pages/assets/master.js:21:18
    at file:///C:/aPORTABLE%20Programs/html-minifier-gh-pages/assets/master.js:138:3

coreyfarrell:
When the value of a boolean attribute contains a value matching
ignoreCustomFragments we html-minifier should not remove the value even
if collapseBooleanAttributes is set.

Fixes #931

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -595,6 +595,7 @@ function buildAttr(normalized, hasUnarySlash, options, isLast, uidAttr) {
   }
 
   if (typeof attrValue === 'undefined' || options.collapseBooleanAttributes &&
+      !~attrValue.indexOf(uidAttr) &&
       isBooleanAttribute(attrName.toLowerCase(), attrValue.toLowerCase())) {
     attrFragment = attrName;
     if (!isLast) {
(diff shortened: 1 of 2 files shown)


alexlamsl:
https://github.com/kangax/html-minifier/issues/931#issuecomment-393664877

PR #954, opened 2018-08-20
Add command line usage information

JayHoltslander:


README.md
```diff
@@ -136,7 +136,9 @@ npm link .
 
 ## Usage
 
-For command line usage please see `html-minifier --help`
+Note that almost all options are disabled by default. For command line usage please see `html-minifier --help` for a list of available options. Experiment and find what works best for you and your project.
+
+* **Sample command line:** ``html-minifier --collapse-whitespace --remove-comments --remove-optional-tags --remove-redundant-attributes --remove-script-type-attributes --remove-tag-whitespace --use-short-doctype --minify-css true --minify-js true``
```
alexlamsl, line 141: `--minify-css --minify-js` would do, i.e. lose those `true`s as they are the defaults.

alexlamsl (changes requested):
Thanks for your contribution!

Just one minor nit & we will be good to go.

README.md
DamonHD: It wasn't at all obvious to me how to supply (non-default / any) arguments, so I have have never used any with the switches so far, so maybe let this stand to give an extra clue to slow people such as me?

alexlamsl: Sure thing.

alexlamsl (approved):

PR #955, opened 2018-08-21
Throw CleanCSS error

### LINKED ISSUE #953: CleanCSS errors are ignored
[CleanCSS minify method returns an array of errors](https://github.com/jakubpawlowicz/clean-css#minify-method) and it's ignored, should be logged.

https://github.com/kangax/html-minifier/blob/b61dfccfd557167bd97db13091c8fdf7f0f9652a/src/htmlminifier.js#L668

felipemsantana:
Fixes #953

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -665,7 +665,11 @@ function processOptions(values) {
           return prefix + quote + options.minifyURLs(url) + quote + suffix;
         });
         try {
-          return unwrapCSS(new CleanCSS(value).minify(wrapCSS(text, type)).styles, type);
+          var cleanCssOutput = new CleanCSS(value).minify(wrapCSS(text, type));
+          if (cleanCssOutput.errors.length > 0) {
+            throw new Error(cleanCssOutput.errors[0]);
+          }
+          return unwrapCSS(cleanCssOutput.styles, type);
         }
         catch (err) {
           options.log(err);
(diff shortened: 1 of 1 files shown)


src/htmlminifier.js
```diff
@@ -665,7 +665,11 @@ function processOptions(values) {
           return prefix + quote + options.minifyURLs(url) + quote + suffix;
         });
         try {
-          return unwrapCSS(new CleanCSS(value).minify(wrapCSS(text, type)).styles, type);
+          var cleanCssOutput = new CleanCSS(value).minify(wrapCSS(text, type));
+          if (cleanCssOutput.errors.length > 0) {
+            throw new Error(cleanCssOutput.errors[0]);
+          }
+          return unwrapCSS(cleanCssOutput.styles, type);
```
alexlamsl: Does `clean-css` throw anymore?

If not, let's get rid of the whole try-catch, plus that way we get to log all the errors not just the first one in the list.

alexlamsl (changes requested):
Thanks for looking into this :+1:

PTAL at suggestion to simplify error handling.

src/htmlminifier.js
felipemsantana: Only for unhandled exceptions, but most of them should be handled.
Done in https://github.com/kangax/html-minifier/pull/955/commits/56e40e4f726ec2e7599377680e059a2ffdb06ffa @alexlamsl

alexlamsl (approved):

PR #965, opened 2018-09-23
Prevent data loss when using `decodeEntities`

### LINKED ISSUE #964: decodeEntities unsafely decodes &amp; when it results in a legacy character reference
The `decodeEntities` option decodes `&amp;` even in cases when doing so changes the meaning (and rendering) of the resulting HTML.

Input HTML:

```html
<pre><code>Label current;
// Load effective address of current instruction into rcx.
__ leaq(rcx, Operand(&amp;current));
__ bind(&amp;current);
</code></pre>
```

With `decodeEntities: true`, this gets minified into:

```html
<pre><code>Label current;
// Load effective address of current instruction into rcx.
__ leaq(rcx, Operand(&current));
__ bind(&current);
</code></pre>
```

Note how `&amp;current` (which renders as “&current”) is changed into `&current` (which renders as “¤t”). This should not happen since `&curren` (yes, even without the trailing `;`) is a character reference for `¤`, and so turning the `&amp;` into just `&` changes the rendering/meaning.

The full list of such entities that do not require a closing `;` (like `&curren`) is here: https://github.com/mathiasbynens/he/blob/master/data/decode-map-legacy.json ([generated](https://github.com/mathiasbynens/he/blob/master/scripts/process-data.js) from the HTML Standard)

mathiasbynens:
Without this patch, the `decodeEntities` option decodes e.g. `&amp;` even in cases when doing so changes the meaning (and rendering) of the resulting HTML.

Fixes #964.


tests/minifier.js
```diff
@@ -3353,6 +3353,11 @@ QUnit.test('decode entity characters', function(assert) {
   assert.equal(minify(input, { decodeEntities: false }), input);
   assert.equal(minify(input, { decodeEntities: true }), input);
 
+  // https://github.com/kangax/html-minifier/issues/964
+  input = '&amp;xxx; &amp;xxx &ampthorn; &ampthorn &ampcurren;t &ampcurrent';
+  output = '&ampxxx; &xxx &ampthorn; &ampthorn &ampcurren;t &ampcurrent';
```
mathiasbynens, line 3358: Without this patch, the output is:

```js
'&ampxxx; &xxx &ampampthorn; &thorn &ampampcurren;t &current'
```

…which renders as:

```
&xxx; &xxx &ampthorn; þ &ampcurren;t ¤t
```

…whereas the input HTML renders as:

```
&xxx; &xxx &thorn; &thorn &curren;t &current
```

mathiasbynens:
One of the CI builds failed with… no output whatsoever: https://travis-ci.org/kangax/html-minifier/jobs/432193341 I don’t think that’s related to the patch.

PR #966, opened 2018-09-25
Omit `nomodule` attribute value as it's boolean

shinnn:
### Before

```js
const {minify} = require('html-minifier');

minify('<script nomodule="nomodule"></script>', {collapseBooleanAttributes: true});
//=> '<script nomodule="nomodule"></script>'
```

### After

```js
const {minify} = require('html-minifier');

minify('<script nomodule="nomodule"></script>', {collapseBooleanAttributes: true});
//=> '<script nomodule></script>'
```

------

ref. specification of `nomodule` attribute:  https://html.spec.whatwg.org/multipage/scripting.html#attr-script-nomodule

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -199,7 +199,7 @@ function isStyleSheet(tag, attrs) {
   return true;
 }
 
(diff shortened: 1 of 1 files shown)


shinnn:
Could anyone review this?

alexlamsl:
Please add a test case (around [here](https://github.com/kangax/html-minifier/blob/v3.5.21/tests/minifier.js#L1425-L1428)) to make sure your change is working as expected.

PR #970, opened 2018-10-05
Use <!doctypehtml> (without the space) where possible

### LINKED ISSUE #969: Minify <!doctype html> into <!doctypehtml>
@krijnhoetmer pointed out that per https://html.spec.whatwg.org/multipage/parsing.html#parse-error-missing-whitespace-before-doctype-name, `<!doctypehtml>` behaves just like `<!doctype html>`. It’s a parse error and thus technically invalid HTML, but it still successfully triggers standards mode!

> This error occurs if the parser encounters a DOCTYPE whose "DOCTYPE" keyword and name are not separated by ASCII whitespace. In this case the parser behaves as if ASCII whitespace is present.

mathiasbynens:
When both the `useShortDoctype` and the `removeTagWhitespace` options are enabled, we can remove the space from the doctype altogether.

It’s invalid HTML (which is already expected in the output with the `removeTagWhitespace` option enabled), but it still triggers standards mode.

https://html.spec.whatwg.org/multipage/parsing.html#parse-error-missing-whitespace-before-doctype-name
https://twitter.com/mathias/status/1048098677703753729

Closes #969.

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -1235,7 +1235,9 @@ function minify(value, options, partialMarkup) {
       buffer.push(text);
     },
     doctype: function(doctype) {
-      buffer.push(options.useShortDoctype ? '<!doctype html>' : collapseWhitespaceAll(doctype));
+      buffer.push(options.useShortDoctype ? '<!doctype' +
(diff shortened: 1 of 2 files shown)


kangax:
this is nuts

PR #971, opened 2018-10-12
Html-minifier: Removed Phantomjs dependency

ossdev07:
PhantomJS is an unmaintained project, removed phantomjs dependency from package.json and Gruntfile.js.
dist/htmlminifier.js , dist/htmlminifier.min.js and  tests/index.html files got updated whenever we run the package.

Signed-off-by: ossdev07 <ossdev@puresoftware.com>


ossdev07:
Ping @alexlamsl @kangax 

Please have a look at this PR.
It will be a delight to see phantomjs removed from html-minifier.

rhenwood-arm:
I would also like to see PhantomJS removed. Can you provide feedback on this PR @kangax @mathiasbynens ?

mathiasbynens:
I wasn’t yet a maintainer when this was added, so I don’t feel confident taking action on this PR. IMHO, if we want to drop this dependency, we should make sure we have an adequate replacement so as not to lose out on test coverage.

kangax:
I haven't worked on html-minifier in a while. Don't recall exactly what we used phantom for. I'm all for removing it, but what would be a good replacement? Puppeteer?

ossdev07:
Hi @kangax @mathiasbynens 

Yes, we can replace phantomjs with puppeteer or chromium.
What do you suggest, how should we move forward with this.

alexlamsl:
It is used by `npm test` to test the variant of `html-minifier` for web browsers.

You cannot just remove **half** of our testing just because you don't feel like it. And users won't even install anything inside `devDependencies` anyway.

ossdev07:
Sorry to bother you again @alexlamsl 

I agree that removing the phantomjs is not a good idea, but as @mathiasbynens and @kangax were mentioning we can replace the phantomjs with some alternative like chrome headless. 

Please share your views, will it be fine for the package if we replace phantomjs with chrome headless because phantomjs is an unmaintained project and chrome headless is a good alternative for it.

kangax:
I would be in favor of removing it since Phantom is basically dead (as far as I know) and it'd be nice to use modern tool like puppeteer that would likely be faster and more powerful. To clarify, we would have to port everything to Puppeteer, not just remove Phantom, slashing our test coverage.

alexlamsl:
I'm not against replacing our current browser testing framework either, just not removing it.

Personally, I won't be sparing any cycles on this though.

PR #975, opened 2018-10-28
Also minify CSPs (fixes #947)

### LINKED ISSUE #947: Please minify Content Security Policy
Current result:

image

ENT8R:
I hope everything is OK. (I didn't take a look into the whole codebase and don't know if this would be the right place to add it...)


XhmikosR:
Please fix the test failures and rebase against upstream/master.

ENT8R:
I fixed the indentation issues with https://github.com/kangax/html-minifier/pull/975/commits/fa17d5c87058875d669fba786f3f100a5124465f
Github IDE screwed something up there... Sorry for that.

XhmikosR:
Looks good now. One thing I don't know though because I haven't dug deep into the codebase myself either, is if we need to lowercase tags, their names and/or their values before the comparisons.

alexlamsl:
HTTP headers are case insensitive, plus the follow two MDN links listed `<meta>` usage using different case schemes:
https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta

So I'd say we need to do a lower case comparison here. Also, can we get a test case or two for this new feature?

Just add a new section at the end of [`tests/minifier.js`](https://github.com/kangax/html-minifier/blob/gh-pages/tests/minifier.js) like so:
```js
QUnit.test('minify Content-Security-Policy', function(assert) {
  assert.equal(minify(...), ...);
  ...
});
```

ENT8R:
So now there are also some test cases and a case insensitive check with https://github.com/kangax/html-minifier/pull/975/commits/e06608461948f191ea25484d61a7cd551344e1e9

alexlamsl:
Changes LGTM &minus; now if I can just figure out how to get the Cursed Integration job to start for this PR...

alexlamsl (approved):

PR #985, opened 2018-12-28
Show help if no arguments supplied 

jamesgeorge007:
Presently if the user fires in `html-minifier` without supplying any of the options, nothing is gonna happen. It is good to have the help showing up in that case 🤔


DamonHD:
I counter-suggest that plenty of well-understood applications designed to work in a pipeline behave exactly like that and for good reason, all the way up from the humble cat, sed, awk, grep, etc...  And this is in many ways a glorified filter too.

Rgds

Damon

jamesgeorge007:
I think it is a good idea to have this since the user would get an idea of the various commands available for the very first time :thinking:

DamonHD:
It's fairly well understood that if a (filter) command is silent with no args then you try it with -help or --help.  Else **every single time** you use it as a filter you need to add arguments and clutter to suppress that (unwanted) output.

For non-filter-pipeline commands your case is stronger.

Rgds

Damon

jamesgeorge007:
@DamonHD  Most of the CLI tools these days ship with this option. I guess it is good to have despite the thoughts above :thinking:

PR #993, opened 2019-01-28
Move to Terser (allows ES6 to be minified)

C-Lodder:
Uglify-JS doesn't support ES6 and Uglify-ES is no longer maintained.

Terser is a fork of Uglify-ES that maintains the API, therefore this won't break B/C.

With this PR, you'll noticed that ES6 is now minified properly, as opposed to being ignored


C-Lodder:
@alexlamsl I've reopened this. What are you thoughts?

The main reason I've done this is because some users, like myself use a small inline script at thew bottom of the page to lazy load CSS/JS files. If that inline script is ES6, then it won't get minified.

alexlamsl:
There is already a way to supply custom JavaScript minification to `html-minifier`, so I have no desire to change the default to anything untested.

C-Lodder:
@alexlamsl - Do you mean inject pre-minified code? Or do you mean minify inline JS inside a HTML file?

If the later, then this doesn't work with ES6. Terser is simply a fork of Uglify-JS (which is no longer maintained)

alexlamsl:
> Uglify-JS (which is no longer maintained)

I maintain `uglify-js` - no sure why you are making such a false claim.

C-Lodder:
@alexlamsl My bad, I meant Uglify-ES, not JS. Either way it doesn't seem to minify ES6

PR #1001, opened 2019-03-21
fix: <?php ?>0 in script will throw an error

zhihuahuang:
If you use `minifyJS` options, a number after custom fragments in script will throw an error.

Like this:

```
let html = `<script>var value = "<?php ?>+<?php ?>0";</script>`;

minify(html, {
    minifyJS: true,
});
```

It will be throw a `TypeError`.

```
TypeError: Cannot read property '0' of undefined
    at html-minifier/src/htmlminifier.js:1266:52
```

I found that the cause is the regular expression `uidAttr + ([0-9]+)`. It will match the extra numbers. So I added parentheses to the regular expression.


alexlamsl:
Please add some test cases to demonstrate this fix. Also may I suggest `uidAttr + ([0-9]+?) + uidAttr` instead of adding parenthesis for better resilience.

Issue #974, opened 2018-10-24
node benchmark.js is broken


mathiasbynens:
Here's what I eventually get when running `node benchmark.js`:

```
[...]
Error: ENOENT: no such file or directory, open 'benchmarks/generated/New York Times.minifier.html'
```

mathiasbynens:
This happens because nytimes.com has the following in their HTML:

```html
<!––
         0000000                         000        0000000
       111111111      11111111100          000      111111111
       00000        111111111111111111      00000      000000
       000        1111111111111111111111111100000         000
       000        1111       1111111111111111100          000
       000         11       0     1111111100              000
       000          1      00             1               000
       000               00      00       1               000
       000             000    00000       1               000
    00000            0000  00000000       1                00000
  11111            000 00    000000      000                 11111
    00000          0000      000000     00000              00000
       000        10000      000000      000              0000
       000        00000      000000       1               000
       000        000000     10000        1     0         000
       000        1000000 00              1    00         000
       000         1111111                1 0000          000
       000          1111111100           000000           000
       0000          111111111111111110000000            0000
       111111111        111111111111100000          111111111
         0000000              00000000              0000000
  
      NYTimes.com: All the code that's fit to printf()
      We're hiring: https://nytimes.wd5.myworkdayjobs.com/Tech
   -->
```

Note that `<!––` contains two U+2013 characters as opposed to just simple hyphens (i.e. `<!--`).

mathiasbynens:
https://twitter.com/mathias/status/1055130976370745344

michaelficarra:
I'm sure they were just following the style guide.

mathiasbynens:
The fix is not yet live on nytimes.com, so for https://github.com/kangax/html-minifier/commit/5494091bcc150b30db71749978cfb32b368111c7 I cheated by temporarily replacing the URL on this line: https://github.com/kangax/html-minifier/blob/5494091bcc150b30db71749978cfb32b368111c7/benchmarks/index.json#L11 with a copy of the HTML response where the two U+2013 characters are replaced with regular hyphens.

PR #1004, opened 2019-03-22
implement `continueOnParseError`

alexlamsl:
Allow parser to recover and treat invalid characters as text.

Previously, all invalid input would abort without any minification.

There are many issues aside from #974 that users have filed over the years &minus; this PR should allow them to continue without resorting to fix their own broken HTML inputs.


sabriidelpoy:
@alexlamsl This feature is great! When are you going to publish a new version with this included?

PR #1005, opened 2019-03-22
Version 3.6.0

alexlamsl:


XhmikosR:
I think this needs to be v4.0.0 due to the minimum Node.js version change.

That being said, we could also wait for the ES6 change, or not.

/CC @kangax

alexlamsl:
I was hoping to get the current changes out quickly.

If you want 4.0 with ES6 stuff, let's back out the Node.js version changes, release 3.6.0, then you can do whatever you wish going forward :wink:

XhmikosR:
Nah, just release this as 4.0, it's fine. Numbers are infinite :P

XhmikosR:
Also, let's wait for @kangax and @mathiasbynens  to chime in.

mathiasbynens:
v4.0.0 sounds good to me!

alexlamsl:
Much as I would like to update the version number, right now everything is in lock-down because of the bloody Branch Protection thingy combined with non-working Travis CI build for PRs.

XhmikosR:
Travis works fine as you can see by the other PRs. Make sure you rebase
your branch which is why I suggested for this option to be enabled so that
we have a button.

Also this should be marked v4.0.0 not 3.6.0.

alexlamsl:
I have tried it on other PRs. If you look at them closer they only worked while we were still on `travis-ci.org`, and has stopped working ever since we've migrated to `travis-ci.com` - I even created a new PR to test this (#1014) but as you can see only the branch job was triggered, the PR one which prevents me from merging PRs is never run.

alexlamsl:
As you can see even your own Pull Request (#1009) isn't getting through 😅 pr-ci-not-working

(I've referring to the `required` one which isn't even being run, not the failed test above)

XhmikosR:
Then it's something else wrong because it worked fine.

XhmikosR:
We might as well drop Travis...

XhmikosR:
Must have been the Travis org to com switch.

alexlamsl:
@XhmikosR thanks to @kangax the thingy now works for local branches, so at least we have a somewhat functional way to move forward for now 😉

PR #1006, opened 2019-03-22
Update .travis.yml

XhmikosR:
Remove unneeded stuff and add Node.js 10.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,29 +4,38 @@ addons:
       - ubuntu-toolchain-r-test
     packages:
       - libstdc++-4.9-dev
+
 branches:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,29 +4,38 @@ addons:
       - ubuntu-toolchain-r-test
     packages:
       - libstdc++-4.9-dev
+
 branches:
   only:
     - gh-pages
+
 language: generic
```
XhmikosR, line 12: Still not sure about this choice. Seems wrong, but I haven't looked into it more.

kangax (approved):

PR #1008, opened 2019-03-22
 Drop Node.js < 6 support. 

### LINKED ISSUE #1007: Drop Node <6 support now
This is insane. There is absolutely no logical reason to keep v4 supported.

It's been EOL for so long already and by keep supporting it, we actually do things worse since the users are open to so many vulnerabilities.

I'm gonna make a PR and IMO this, we shouldn't even discuss it.

XhmikosR:
Should be merged after #1006 is merged.

More ES6-ify patches will come later, in a separate PR.

Fixes #1007

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ matrix:
 
 env:
   matrix:
-    - NODEJS_VER=node/4
     - NODEJS_VER=node/6
     - NODEJS_VER=node/8
     - NODEJS_VER=node/10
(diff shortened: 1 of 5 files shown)


kangax:
Is it safe to drop Node 6 as well?

kangax (approved):

XhmikosR:
The reason I didn't drop 6, is that it's still supported until next month.

Better play it safe here.

PR #1009, opened 2019-03-23
[WIP] ES6-ify codebase.

XhmikosR:
So far I need to add babel in browserify, any help welcome. And then see if tests pass.

Also, I plan to move to xo for linting because it's much better than what we have now.


XhmikosR:
So far I need to add babel in browserify, any help welcome. And then see if tests pass.

alexlamsl:
May I ask why we need this? 🤔

I mean, the existing code works and any ES6 supported platform will happily run it.

OTOH, any old JavaScript web browsers will now have trouble running the updated code base.

kangax:
My understanding is that @XhmikosR is planning to add Babel compile step. 

I'm all for ES6+ codebase, which is used in all of my work projects, and without which I wouldn't start any new ones. 

However, I'm also ok with just keeping this as is. IMO, the codebase seems stable and doesn't change often enough. Are all those consts and template literals really worth adding an extra compile step? I dunno. 

The Babel compile step is obviously so ubiquitous these days that it could barely be considered a downside of any kind. It would, however, add a bit more complexity to the infrastructure of the project.

All in all — I don't see much need in this, but if that's what y'all prefer, I think that's also fine :)

alexlamsl:
I can understand the desire to use ES6 for new code bases &minus; we would have structured it quite differently due to the new language features if we were to build this from the ground up anyway.

kangax:
It's up to you guys. I don't have time to work on this right now but if you want to revamp, go for it.

XhmikosR:
My reasoning is simple. It just makes more sense to write in ES6+ than legacy.

XhmikosR:
On a related note, that custom uglify-js logic in Gruntfile.js is what breaks things, because it runs before the babel transform and as we know uglify doesn't support ES6.

What was the reasoning about this logic instead of just concatenating the files?

alexlamsl:
By custom logic I guess you mean the packaging required for `uglify-js` &minus; all it does is grab its own source files (which are all ES5) and produce something that can be consumed by `require()`

It shouldn't fail due to ES6 because it doesn't take any outside code as input.

XhmikosR:
I actually meant https://github.com/kangax/html-minifier/blob/gh-pages/Gruntfile.js#L37

I'll dive into this later, but as far as I can tell, we just use rollup for creating the bundle.

alexlamsl:
> I actually meant https://github.com/kangax/html-minifier/blob/gh-pages/Gruntfile.js#L37

Yes, we are talking about the same thing here. Something that takes in guaranteed ES5 input and gives pure ES5 output shouldn't cause any issues within ES6 stuff.

XhmikosR:
Yeah, after playing around yesterday a bit with it, it seems the babel transforms aren't applied.

I'll find a solution later, but I will go the rollup way probably.

PR #1056, opened 2020-01-09
Fixing exceptions when attribute contains no value

Chopinsky:
This PR is to address the exception being thrown if the target html file contains `link` tags whose `type` attribute is defined but without a value. 

For example, when using the plugin with webpack, encountering the following `link` tag: 

`<link href="https://somecdn.com/favicon/favicon.png" rel="icon" type>` 

in a html file will crash the build process because of this exception: `TypeError: Cannot read property 'toLowerCase' of undefined`.

With this fix, we will treat the undefined attribute value as a blank string value, i.e. it will be equivalent as the `link` tag below:

`<link href="https://somecdn.com/favicon/favicon.png" rel="icon" type="">`.

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -183,11 +183,13 @@ function isExecutableScript(tag, attrs) {
 
 function isStyleLinkTypeAttribute(attrValue) {
   attrValue = trimWhitespace(attrValue);
-  if (!attrValue) {
-    return true;
+
+  if (typeof attrValue !== 'string') {
+    attrValue = '';
+  } else {
+    attrValue = attrValue.toLowerCase();
   }
   
-  attrValue = attrValue.toLowerCase();
   return attrValue === '' || attrValue === 'text/css';
 }
 
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -184,7 +184,7 @@ function isExecutableScript(tag, attrs) {
 function isStyleLinkTypeAttribute(attrValue) {
   attrValue = trimWhitespace(attrValue);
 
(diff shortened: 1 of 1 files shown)


src/htmlminifier.js
```diff
@@ -182,7 +182,14 @@ function isExecutableScript(tag, attrs) {
 }
 
 function isStyleLinkTypeAttribute(attrValue) {
-  attrValue = trimWhitespace(attrValue).toLowerCase();
+  attrValue = trimWhitespace(attrValue);
+
+  if (typeof attrValue !== 'string') {
```
DanielRuf: Wouldn't be `typeof attrValue !== 'undefined'` be better?

Because the quotes are often optional (unquoted attribute values) so this would change `val=integer` to `val=''`.

DanielRuf (changes requested):

src/htmlminifier.js
DanielRuf: https://dev.w3.org/html5/html-author/Overview.src.html

> There are four slightly different syntaxes that may be used for attributes in HTML: empty, unquoted, single-quoted and double-quoted. All four syntaxes may be used in the HTML syntax, depending on what is needed for each specific attribute. However, in the XHTML syntax, attribute values must always be quoted using either single or double quotes.

https://mathiasbynens.be/notes/unquoted-attribute-values

Only undefined / empty values should be treated this way.

Chopinsky: I've made the update to only handle the 'undefined' values here.

lucalves (approved):

PR #1071, opened 2020-08-10
fix collapseWhitespace for <output> tag

### LINKED ISSUE #1070: collapseWhitespace remove the space around the <output> tag
**how to reproduce:**

1. create example html file "input.html":
```html
<!-- It's not ok! Removes space before and after number 23 -->
<p>Before output tag text <output>23</output> after output tag text.</p>
<!-- It's ok -->
<p>Before b tag text <b>23</b> after b tag text.</p>
```

2. run with collapseWhitespace parameter
```npx html-minifier --collapse-whitespace input.html -o output.html```

**html result "output.html":**
```html
<!-- It's not ok! Removes space before and after number 23 --><p>Before output tag text<output>23</output>after output tag text.</p><!-- It's ok --><p>Before b tag text <b>23</b> after b tag text.</p>
```

**reason: the "output" is not a inline tag, but is often used as a inline (for javascript).**

epexa:
Fixes #1070

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -64,7 +64,7 @@ function collapseWhitespace(str, options, trimLeft, trimRight, collapseAll) {
 
 var createMapFromString = utils.createMapFromString;
 // non-empty tags that will maintain whitespace around them
-var inlineTags = createMapFromString('a,abbr,acronym,b,bdi,bdo,big,button,cite,code,del,dfn,em,font,i,ins,kbd,label,mark,math,nobr,object,q,rp,rt,rtc,ruby,s,samp,select,small,span,strike,strong,sub,sup,svg,textarea,time,tt,u,var');
+var inlineTags = createMapFromString('a,abbr,acronym,b,bdi,bdo,big,button,cite,code,del,dfn,em,font,i,ins,kbd,label,mark,math,nobr,object,output,q,rp,rt,rtc,ruby,s,samp,select,small,span,strike,strong,sub,sup,svg,textarea,time,tt,u,var');
 // non-empty tags that will maintain whitespace within them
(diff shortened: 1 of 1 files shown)


DanielRuf:
Awesome. Can you provide a PR for this at https://github.com/DanielRuf/html-minifier-terser? That would be really great.

DanielRuf:
Also I think it makes sense to add some test for this. Can you give this a try or do you need help with this?

XhmikosR:
@epexa yeah, see if you can add a test case please.

@kangax @DanielRuf might be worth joining forces in one repo. I personally do use terser for my JS files, but worst case we could make the JS minifier configurable?

DanielRuf:
> @kangax @DanielRuf might be worth joining forces in one repo.

Sounds definitely good. But personally I have minimized the time that I invest in opensource and I have onboarded two new maintainers from tje JS and webpack community (see https://github.com/terser/html-minifier-terser/issues/79). To be honest, I'm struggling with depressions and somewhat feel burned out. So please do not expect too much from my side for now.

In general it would be probably better to ask the new maintainers and I can try to do some reviews or express my opinion in some cases.

Joining forces, bringing html-minifier forward and work together instead of against each other would be ideal in the opensource community and something that I have always tried in other projects in the past (for https://github.com/materializecss/materialize I invited the original maintainers but unfortunately they never accepted the invitations, but also life is more important probably).

I've chosen the terser community as new home since I thought this would be a good place and I've learned from other projects that there are many problems when the owner is not an org but a normal user (I have worked on projects where the owners left this world and there were no more changes done), especially regarding the bus / truck factor and other things. I'm open to share projects and invite others like I did with https://github.com/website-checks/website-checks.

PR #1073, opened 2020-08-24
Added in copy button

Yash-Singh1:
# Overview
Added in a button called `Copy Result` with an ID of `copy-btn` allowing you to copy the result of the minification process.
# List of Updates
Here is a list of the updates:
* Added in a button inside `index.html` with the same CSS as the earlier `minify-btn`
* Enabled the `copy-btn` on click of the `minify-btn`
* Copied the result on click of the `copy-btn`
# Testing
To test my changes inside a development environment, go to: https://yash-singh1.github.io/html-minifier/


assets/master.js
```diff
@@ -104,6 +105,19 @@
       byId('stats').innerHTML = '<span class="failure">' + escapeHTML(err) + '</span>';
       byId('minify-btn').disabled = false;
     });
+    byId('copy-btn').onclick = function() {
+      let bgColorButton = byId('copy-btn').style.backgroundColor;
+      let copyText = byId('output');
+      copyText.select();
+      copyText.setSelectionRange(0, 99999);
```
DanielRuf: This could lead to problems. Why not simply fetch the value and copy it with `window.Clipboard.writeText()`?

See https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText

assets/master.js
```diff
@@ -104,6 +105,19 @@
       byId('stats').innerHTML = '<span class="failure">' + escapeHTML(err) + '</span>';
       byId('minify-btn').disabled = false;
     });
+    byId('copy-btn').onclick = function() {
+      let bgColorButton = byId('copy-btn').style.backgroundColor;
```
DanielRuf: Can / should be done with a class.

assets/master.js
```diff
@@ -104,6 +105,19 @@
...
+      let copyText = byId('output');
+      copyText.select();
+      copyText.setSelectionRange(0, 99999);
+      document.execCommand("copy");
+      byId('copy-btn').innerHTML = 'Copied!'
+      byId('copy-btn').style.backgroundColor = 'green';
+      setTimeout(function() {
+        byId('copy-btn').innerHTML = 'Copy Result';
```
DanielRuf: Via CSS would be probably better.

assets/master.js
Yash-Singh1: Sure, will make a commit doing that. What should be the response when the access is denied to write to the clipboard? Should we respond with `alert("Access denied to the clipboard")`?

DanielRuf: > Should we respond with `alert("Access denied to the clipboard")`?

Sounds good.

Yash-Singh1: Added this change in latest commit: b6249efc6e084ed7aa46ab511df9f811b36f4054

assets/master.js
Yash-Singh1: @DanielRuf I didn't get it, can you elaborate?

DanielRuf: Don't set (inline) styles with JS, set a class with JS and apply the needed styles via style tag. That's is better to maintain.

Yash-Singh1: Sure, added it in!

assets/master.js
Yash-Singh1: Is this review just a duplicate of the previous one asking for CSS, or do you mean something different. I think c5d03e9 adds in using the `assets/master.css` instead of just plain editing through JS.

DanielRuf: It's about the same.

DanielRuf: You can ignore this as the change is outdated because of your changes. So it's fine now.

Bildschirmfoto 2020-09-11 um 00 23 35

Yash-Singh1:
Any updates on this pull request, or is it good to go?

DanielRuf:
LGTM

Yash-Singh1:
So, will @kangax merge it?

kangax:
hm, Travis CI doesn't let me merge, says some checks haven't completed yet

Yash-Singh1:
@kangax Looks like it says that here too:
Screenshot from 2020-09-22 17-53-34
Look at this:
https://travis-ci.community/t/known-issue-travis-ci-reports-expected-waiting-for-status-to-be-reported-on-the-github-status-api-but-the-status-never-arrives/1154
I think it is a fix. Looks like this was an issue: https://github.com/travis-ci/travis-ci/issues/10187
Here is another write-up someone wrote on the fix: https://github.com/travis-ci/travis-ci/issues/10204
I think the solution is just to make sure that you have access to merge this pull request even when the tests haven't passed. This should be enabled in scenarios in which you want to force merge. I ran the tests manually, here is what appeared:
Screenshot from 2020-09-22 18-03-13
If you want a close up of the results:
Screenshot from 2020-09-22 18-03-54
Looks like all the tests worked manually. :+1:

Yash-Singh1:
Since `html-minifier` now uses Gh Actions, can this be merged?

kangax:
yeah, I'm ok with it

@XhmikosR any objections?

XhmikosR:
Yeah, this shouldn't land as is. `innerHTML` is meaningless when we only need to change the text.

Also, I believe we should keep the initial text instead of re-adding it.

DanielRuf:
> Yeah, this shouldn't land as is. `innerHTML` is meaningless when we only need to change the text.

You are right, `innerText` should be sufficient then.

Issue #860, opened 2017-10-17
default true CLI options can't be disabled ("include-auto-generated-tags")


bulk88:
given file
```
<html><head><meta name="mobileoptimized" content="0"></head><body>
<a href="#1">&nbsp;1&nbsp;</a> 
<a href="#2">&nbsp;2&nbsp;</a> 
<a href="#3">&nbsp;3&nbsp;</a> 
<a href="#4">&nbsp;4&nbsp;</a> 
<a href="#5">&nbsp;5&nbsp;</a>
<br><br>
<a name="1">1: <a href="#A">A</a><br>
<a name="2">2: <a href="#B">B</a><br>
<a name="3">3: <a href="#C">C</a><br>
<a name="4">4: <a href="#D">D</a><br>
<a name="5">5: <a href="#E">E</a><br>
<br><br>
<a name="A"><a href="s.htm#A1">Page 1</a><br>
<a name="B"><a href="s.htm#A2">Page 2</a><br>
<a name="C"><a href="s.htm#A3">Page 3</a><br>
<a name="D"><a href="s.htm#A4">Page 4</a><br>
<a name="E"><a href="s.htm#A5">Page 5</a><br>
</body></html>
```
which has target anchors with no contents and no closing tag. No closing tag is on purpose for space optimization, since HTML prohibits an A tag in an A tag and implicitly closes the last one when it encounters an opening A tag. I am using the following line to minify this HTML file.

```
html-minifier --collapse-boolean-attributes --collapse-whitespace  --html5 --remove-attribute-quotes --remove-comments  --remove-empty-attributes --remove-optional-tags  --remove-redundant-attributes --remove-script-type-attributes  --remove-style-link-type-attributes --remove-tag-whitespace  --sort-attributes --sort-class-name --trim-custom-fragments  --use-short-doctype --minify-js -o testm.htm test.htm
```
gives 

[...]

PR #1090, opened 2021-04-14
fix config file: "html5" and "includeAutoGeneratedTags" were ignored

### LINKED ISSUE #1045: includeAutoGeneratedTags not working properly
Here is a test case:
```
<!DOCTYPE html>
<html>
  <head>
  </head>
  <body>
    <div>
      <header> 
      </header>
```

When I minify it with `includeAutoGeneratedTags` set to `false`, I get the following output `<!doctype html><div><header></header></div>`. As you can see, a closing `</div>` is added. It is a huge problem when minifying template files.

In case it is relevant, here are all the options which I have (although I did try to change them and they did not influence current behaviour with `div`):
```
{
  "caseSensitive": false,
  "collapseBooleanAttributes": true,
  "collapseInlineTagWhitespace": true,
  "collapseWhitespace": true,
  "conservativeCollapse": false,
  "decodeEntities": true,
  "html5": true,
  "includeAutoGeneratedTags": false,
  "keepClosingSlash": false,
  "maxLineLength": 0,
  "minifyCSS": true,
  "minifyJS": true,
  "preserveLineBreaks": false,
  "preventAttributesEscaping": false,
  "processConditionalComments": true,
  "processScripts": [
    "text/html",
"application/ld+json"
  ],
  "removeAttributeQuotes": false,
  "removeComments": true,
  "removeEmptyAttributes": true,
  "removeEmptyElements": false,
  "removeOptionalTags": true,
  "removeRedundantAttributes": true,
  "removeScriptTypeAttributes": true,
  "removeStyleLinkTypeAttributes": true,
  "removeTagWhitespace": false,
  "sortAttributes": true,
  "sortClassName": true,
  "trimCustomFragments": true,
  "useShortDoctype": true
}
```

I thought that #540 was supposed to add an option to remove such behavior. Is there any other way yo turn off such additions of closing tags?

bulk88:
the fix for #860 enabled disabling these 2 flags from the command line
but also made them uncontrollable from the JSON config file

old commander.js (not 2021 commander.js) assigns boolean true
unconditionally as defaultValue for all --no- arguments, when minifier
fuses command line and JSON config file options together in cli.js's
createOptions(), it sees program["includeAutoGeneratedTags"] = true and
ignored config["includeAutoGeneratedTags"], in my case "= false". Delete
the defaultValue/key from commander.js object before parsing user's ARGV
to solve the bug.

EXTRA NOTES this probably fixes https://github.com/kangax/html-minifier/issues/1045

work around for old/publish html-minifiers is to always put `--no-include-auto-generated-tags` on cmd line in addition to `  "includeAutoGeneratedTags": false` in json config file


DanielRuf:
Would be great to have this PR at https://github.com/terser/html-minifier-terser too =)

XhmikosR:
@bulk88 does this apply to the latest commander too? I'm wondering because the plan is to update all deps to the latest version.

bulk88:
I don't know. commander module from 10 years ago and 2021 commander module have almost no source code in common by now when I debugged this/wrote the patch and there was no attempt at back (bug) compat by commander's authors IIRC.

PR #1099, opened 2021-06-29
README.md: add Gulp usage example without a plugin

brend8c:
I wrote some code that can be used in 'gulp' - gulpfile.js.
There is no need to connect third-party plugins anymore.


DanielRuf:
Would be great to have this PR at https://github.com/terser/html-minifier-terser too =)

brend8c:
> Было бы здорово иметь этот PR на https://github.com/terser/html-minifier-terser тоже =)

Okay, I'll do it after the trip, in a week. I think I can write something like this :)

DanielRuf:
> Okay, I'll do it after the trip, in a week. I think I can write something like this :)

html-minifier-terser is a fork of html-minifier so this should be quite easy (same readme with a few small changes)

no hurry, enjoy your trip

XhmikosR (approved):

PR #1106, opened 2021-10-02
Add CodeQL workflow

XhmikosR:
@kangax can you give me admin rights or remove Travis CI? I will switch to GitHub Actions later.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,34 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches:
+      - gh-pages
+      - "!dependabot/**"
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches:
+      - gh-pages
+  schedule:
+    - cron: "0 0 * * 0"
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v1
+        with:
+          languages: "javascript"
+
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@v1


XhmikosR:
@kangax friendly ping ^^

kangax:
How do we remove it? I checked and it looks like you have "collaborator" access on the repo and there's no other option to select.

XhmikosR:
I planned to also switch branches around and use an Action to deploy. I don't have any other ideas, might be a limitation due to the fact that this repo is a personal one.

So, repo Settings -> Webhooks/Integrations make sure you remove any Travis CI references. Then go to the branch protection rules and adapt it so it's using the Actions from #1107.

kangax:
@XhmikosR ok I think this should do it

XhmikosR:
Are you sure you added the new Actions in the branch protection rules? Because I see I can merge this PR, but no CI has run.

Can you review/approve #1107 and then go to the repo settings to protect the branch again?

XhmikosR:
BTW another handy option is the option to delete HEAD branches :)

XhmikosR:
@kangax can you enable the aforementioned option + the option for the branch to be up to date?

Trying to see how we are going to make the switch from gh-pages to main, and not having the needed rights makes this more complex.

kangax:
@XhmikosR enabled "automatically delete head branches". I don't quite understand if there's anything else I can do in "Require status checks to pass before merging" section. I tried searching in their input field and seeing things like Node 10, Analyze, Page Build, etc. Those are not the same ones as what you added afaics, and when I search for yours nothing comes up.

XhmikosR:
@kangax yeah, that's normal. So, just choose 10, 12, 14, 16. CodeQL might be good too.

XhmikosR:
So, the question is how we solve the 2-3 issues that are valid and are in our source code: https://github.com/kangax/html-minifier/security/code-scanning

* the prototype pollution issue is fixed in the latest clean-css: https://github.com/clean-css/clean-css/commit/cfa26b15fa6ff69583c502fa464b0b9b3ec88f2d
* similarly the useless regular-expression character escape: https://github.com/clean-css/clean-css/commit/9d633cb72b2a9023b1d73055c356a4bb82aebb90
* I'd like to get these sorted somehow: Polynomial regular expression used on uncontrolled data
* the incomplete string escaping or encoding, we need to confirm if it's valid

PR #1109, opened 2021-10-02
Fix/update benchmark

XhmikosR:
`htmlcompressor.com` has an upper limit of ~300KB hence why it fails. We could probably just drop it later.

Also, IMHO we should move the benchmark script into the benchmark folder along with its own package.json and run it on CI (probably with a cron job since we don't want it to run all the time). Otherwise it can break at any time again.

### DIFF
--- a/benchmark.js
+++ b/benchmark.js
@@ -297,7 +297,7 @@ run(fileNames.map(function(fileName) {
 
     function testWillPeavy(done) {
       readText(filePath, function(data) {
-        var options = url.parse('https://www.willpeavy.com/minifier/');
+        var options = url.parse('https://www.willpeavy.com/tools/minifier/');
         options.method = 'POST';
         options.headers = {
           'Content-Type': 'application/x-www-form-urlencoded'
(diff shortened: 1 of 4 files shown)


benchmark.js
```diff
@@ -254,7 +254,7 @@ run(fileNames.map(function(fileName) {
         // Apply Brotli on minified output
         function(done) {
           readBuffer(info.filePath, function(data) {
-            var output = new Buffer(brotli.compress(data, true).buffer);
+            var output = Buffer.from(brotli.compress(data, true).buffer);
```
XhmikosR: Not sure if `.buffer` is needed anymore

benchmark.js
```diff
@@ -121,9 +119,17 @@ function readSize(filePath, callback) {
...
-  fs.createReadStream(inPath).pipe(zlib.createGzip({
-    level: zlib.Z_BEST_COMPRESSION
-  })).pipe(fs.createWriteStream(outPath)).on('finish', callback);
+  fs.createReadStream(inPath)
+    .pipe(zlib.createGzip())
+    .pipe(fs.createWriteStream(outPath))
+    .on('finish', callback);
+}
+
+function brotli(inPath, outPath, callback) {
```
XhmikosR, line 128: This is the same as the `gzip` function except for the method name. If someone wants to deduplicate this more please go ahead.

PR #1112, opened 2021-10-10
useShortDoctype: always add a space after `doctype`

### LINKED ISSUE #1053: <!doctypehtml>
Hi all,

Trying in this tool : https://kangax.github.io/html-minifier/

I tried minimizing the source-code of that same page. The minimized code does not [validate](https://validator.w3.org/nu/#textarea).
Particularily, the `<!DOCTYPE html>` is reduced to `<!doctypehtml>`, which is not okay.

_To reproduce:_
- Copy this minimal HTML code:

```
<!DOCTYPE html>
<html lang="en">
<head>
	<meta charset="UTF-8" />
	<title>Page</title>
</head>
<body>
	<div class="mydiv">Stuff</div>
</body>
</html>
```

- paste it on that same page
- minify it
- try to [validate](https://validator.w3.org/nu/#textarea) the minified code.

**Result:**

Minified code is 
```
<!doctypehtml><html lang=en><meta charset=UTF-8><title>Page</title><div class=mydiv>Stuff</div>
```
Which is not valid:

> Error: Missing space before doctype name.
> At line 1, column 10

**Expectation:**

Minified code should be valid, ie be atleast this:
```
<!doctype html><html lang=en><meta charset=UTF-8><title>Page</title><div class=mydiv>Stuff</div>
```

**End notes**

This is the only error on my minimal HTML code example.
If you try minifying the HTML of the online tool itself, you’ll find another error, most similar.

Appart from that it seems to do the job. Thanks for that all!

XhmikosR:
Fixes #1053 

IMHO removing this one space isn't worth it compared to outputting valid HTML.

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -1255,9 +1255,11 @@ function minify(value, options, partialMarkup) {
       buffer.push(text);
     },
     doctype: function(doctype) {
-      buffer.push(options.useShortDoctype ? '<!doctype' +
-        (options.removeTagWhitespace ? '' : ' ') + 'html>' :
-        collapseWhitespaceAll(doctype));
+      buffer.push(
+        options.useShortDoctype ?
+          '<!doctype html>' :
+          collapseWhitespaceAll(doctype)
+      );
     }
   });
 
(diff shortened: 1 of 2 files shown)


mathiasbynens:
See https://github.com/kangax/html-minifier/pull/970. This is a feature, not a bug!

mathiasbynens:
The space is only omitted if the `removeTagWhitespace` option is enabled, in which case the minified result isn't guaranteed to be valid HTML anyway (with or withour your patch).

XhmikosR:
Yeah, I just had a second thought and it's fine, indeed. Should have checked closer, all good!

PR #1114, opened 2021-10-12
Update remaining major dependencies

### LINKED ISSUE #1111: Update the remaining major dependencies

R4356th:
Resolves #1111 by updating `camel-case`, `commander` and `param-case`. Apparently also closes #985.

Also makes the CLI output a help message instead of running forever (which is the current behaviour) when no argument has been specified.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -27,13 +27,15 @@
 
 'use strict';
 
-var camelCase = require('camel-case');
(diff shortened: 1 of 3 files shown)


R4356th:
@XhmikosR ping! :)

Issue #1135, opened 2022-10-27
CVE-2022-37620/ ReDoS found in htmlminifier.js


secdevlpr26:
A Regular Expression Denial of Service (ReDoS) flaw was found in kangax html-minifier 4.0.0 via the candidate variable in htmlminifier.js.
The ReDoS vulnerability can be mitigated with several best practices described here: [https://snyk.io/blog/redos-and-catastrophic-backtracking/]

michaeljauk:
Is anyone willing to work on this?

blv-raulcatalan:
Any update on this?

rquadling:
One of the lines referred to be the CVE has the following regex `\s+([1-9][0-9]*w|[0-9]+(?:\.[0-9]+)?x)$`

Would changing it to `\s+([1-9][0-9]*w|[0-9]+(?:\.[0-9]+|)x)$` be enough?

`\s+([1-9][0-9]*w|[0-9]+(?=(\.[0-9]+|))\2x)$` is another way that should stop the backtracking (based this upon https://snyk.io/blog/redos-and-catastrophic-backtracking/.

timbomckay:
I stumbled upon a fork of this maintained by terser: https://www.npmjs.com/package/html-minifier-terser

Probably the way to go.

littleblack111:
any updates?

timbomckay:
> any updates?

@littleblack111 as [mentioned above](https://github.com/kangax/html-minifier/issues/1135#issuecomment-1502065386), along with the slew of references to others switching, [html-minifier-terser](https://www.npmjs.com/package/html-minifier-terser) is probably the best solution. Aside from some repo adjustments this solution appears to be fairly abandoned, especially since the [package hasn't been updated in 5 years](https://www.npmjs.com/package/html-minifier), while `html-minified-terser` is maintained by [Terser](https://terser.org).

rquadling:
Now I've just got to remember what project I'm involved in that has this one as an issue!

Hessah95:
any update on this issue?

timbomckay:
> any update on this issue? 

@Hessah95 The comments in this thread point to another solution that's maintained.

ra-dave:
html-minifier-terser latest version (7.2.0) still has the vulnerability?  I just installed it and it says so.

marcoandre1:
[...]

PR #1153, opened 2024-04-23
chore: update dependencies

j9t:
Bumped dependencies by some simple `npm audit fix`. Though according to npm, this fixed 12 of 15 vulnerabilities, 3 persist; left it there for this update.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,5210 @@
 {
   "name": "html-minifier",
-  "version": "4.0.0",
-  "lockfileVersion": 1,
+  "version": "4.0.1",
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "html-minifier",
+      "version": "4.0.1",
+      "license": "MIT",
+      "dependencies": {
+        "camel-case": "^3.0.0",
+        "clean-css": "^5.2.1",
+        "commander": "^2.20.3",
+        "he": "^1.2.0",
+        "param-case": "^2.1.1",
+        "relateurl": "^0.2.7",
+        "uglify-js": "^3.14.2"
+      },
+      "bin": {
+        "html-minifier": "cli.js"
+      },
+      "devDependencies": {
+        "grunt": "^1.4.1",
+        "grunt-browserify": "^6.0.0",
+        "grunt-contrib-uglify": "^5.0.1",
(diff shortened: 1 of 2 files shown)


j9t:
Btw, when testing, I ran into some issues with ESLint. If not already in progress or pointed out elsewhere, should be worth looking into. html-minifier is using a [now deprecated](https://eslint.org/docs/latest/use/configure/language-options-deprecated) way of configuring ESLint; still, the current issues may be treatable by adding and adjusting

```json
"parserOptions": {
  "ecmaVersion": "latest"
},
```

in the config.)

GintasS:
Hi there. Thank you for your work. Do you think your PR fixes this CVE?

https://www.mend.io/vulnerability-database/CVE-2022-37620

DanielRuf:
@GintasS no, since the relevant code (see my comments in the relevant issue) is in this package and not one of its dependencies.

j9t:
Quick side note, I had already taken the silence here and no updates in the slightly more up-to-date [html-minifier-terser](https://github.com/terser/html-minifier-terser) to fork that and spin another version of html-minifier: [html-minifier-next](https://github.com/j9t/html-minifier-next) ([npm](https://www.npmjs.com/package/html-minifier-next)).

It’s already more up-to-date, it works, I use it with 3 projects, but I’m still slow to announce its availability. Sharing if others like to test and contribute 🤝

DanielRuf:
I think the problem still exists there: https://github.com/j9t/html-minifier-next/blob/8b5c3f7f39fd69fbd1d9ab11bc3412f0ec0f04e4/src/htmlminifier.js#L891

See my analysis starting at https://github.com/kangax/html-minifier/issues/1135#issuecomment-2453388023.

So until this vulnerable regular expression is not changed (meaning: gets some upper limits or some other change to prevent this ReDoS), even the fork may be vulnerable (actually not tested by me).

I'm not an active contributor and maintainer anymore, but if you need some input for this just let me know.

j9t:
@DanielRuf, thanks! Not my domain expertise but I’ve thrown some tooling at a PR to, ideally, resolve that issue in html-minifier-next (and inform similar fixes): https://github.com/j9t/html-minifier-next/pull/20. If you like to have a look at provide feedback, it would be appreciated! (Thanks already for the updates here!)

PR #1157, opened 2024-12-01
Minify content within script tag of type JSON (importmap, manifest, JSON-LD, etc.)

### LINKED ISSUE #1132: Any plans to minify content within script tag of type="application/ld+json"?
Sending a big thanks to the team and contributors maintaining this project 👍 

I wanted to ask if there were any plans to minify content within ```<script type="application/ld+json">```?  Below are all the details of my system and how I've implemented your hard work within my projects. Within that script tag, I am using an array. Hoping this helps. Thanks for reviewing and please let me know if I can give any more details that help.

Version
```"html-minifier": "^4.0.0",```

System
```macOS 12.4```

Node
```v14.16.1```

Usage
```html-minifier --input-dir ../public/ --output-dir ../public/ --case-sensitive --collapse-whitespace --remove-comments --remove-redundant-attributes --remove-script-type-attributes --minify-css true --minify-js true --file-ext html```

### LINKED ISSUE #1152: Add support for <script type="speculationrules">
[<script type="speculationrules"> ](https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API) uses a json format and causes an error on uglifyjs

### PR DESCRIPTION
**Pull request by simonewebdesign**
Resolves #1132, resolves #1152, adds support for import maps and other script tags containing JSON.

### DIFF
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -389,7 +389,7 @@ function processScript(text, options, currentAttrs) {
   for (var i = 0, len = currentAttrs.length; i < len; i++) {
     if (currentAttrs[i].name.toLowerCase() === 'type' &&
         currentAttrs[i].value === 'application/ld+json') {
-      return minifyJson(text);
+      return minifyJson(text, options);
     }
 
     if (currentAttrs[i].name.toLowerCase() === 'type' &&
@@ -400,13 +400,12 @@ function processScript(text, options, currentAttrs) {
   return text;
 }
 
-function minifyJson(text) {
+function minifyJson(text, options) {
   try {
-    // Try parsing and re-stringifying the JSON
     return JSON.stringify(JSON.parse(text));
   }
-  catch (e) {
-    // If JSON is malformed, return the original text
+  catch (err) {
+    options.log(err);
     return text;
   }
 }
--- a/src/htmlminifier.js
+++ b/src/htmlminifier.js
@@ -388,11 +388,27 @@ function cleanConditionalComment(comment, options) {
 function processScript(text, options, currentAttrs) {
   for (var i = 0, len = currentAttrs.length; i < len; i++) {
     if (currentAttrs[i].name.toLowerCase() === 'type' &&
+        currentAttrs[i].value === 'application/ld+json') {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/htmlminifier.js [hunk 1]
```diff
@@ -395,6 +400,17 @@ function processScript(text, options, currentAttrs) {
   return text;
 }
 
+function minifyJson(text) {
+  try {
+    // Try parsing and re-stringifying the JSON
+    return JSON.stringify(JSON.parse(text));
+  }
+  catch (e) {
```
humaknlght: Is there a way to log the error instead of suppressing it?  Similar to what is done here: https://github.com/kangax/html-minifier/blob/ab4c9d6e3e11ab207f545728c918895d5cb4c5f8/src/htmlminifier.js#L740C11-L740C28

simonewebdesign: Done! In f63fdbd :)

src/htmlminifier.js
```diff
@@ -387,6 +387,11 @@ function cleanConditionalComment(comment, options) {
 
 function processScript(text, options, currentAttrs) {
   for (var i = 0, len = currentAttrs.length; i < len; i++) {
+    if (currentAttrs[i].name.toLowerCase() === 'type' &&
+        currentAttrs[i].value === 'application/ld+json') {
```
SuperLlama88888: Import maps (`type="importmap"`) are also often used and should be minified. Could we use an array of acceptable types?

SuperLlama88888 (changes requested):

src/htmlminifier.js
simonewebdesign: Great shout! I've implemented that in 25e77e8a6778f0940b02acb78b4f1dc5403794be. It adds support for import maps and makes it much simpler to implement a new format (it will just be a matter of adding it to the array of acceptable types). I've added more types to the list as well, let me know if you think there are any others I've missed!

SuperLlama88888:
@kangax Really sorry for the ping. Could you please take a look at this PR? Tests are included.

humaknlght (approved):

SuperLlama88888 (approved):

PR #1169, opened 2026-03-26
Update deprecation note

j9t:
Tweaking the deprecation note with info on specific updates in HMN and web version.


j9t:
@kangax, feeling free to merge as is—the Node issues don’t seem salvageable at this point.

DanielRuf:
Awesome.
Can someone with the needed permissions deprecate the package on npmjs with the same message?
https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions

Because without a new release (would not recommend to push a new one, since it has been so long since the last one) the readme update is not reflected at https://www.npmjs.com/package/html-minifier and without the `npm deprecate "message"` none of the tools (npm, yarn, pnpm, ...) will show the information to users on install.

https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions#using-the-command-line or via the package setting for the whole package name on npmjs, both ways work (the way via the settings on npmjs is probably easier as there are some visual inputs instead of the CLI).

MichaelDeBoey:
@kangax @j9t Maybe you could also archive the GitHub repo & deprecate the `html-minifier` package in npm?

DamonHD:
Maybe don't, because some of us cannot quickly or easily upgrade to newer NPM etc?

j9t:
> Maybe don't, because some of us cannot quickly or easily upgrade to newer NPM etc?

Archiving the repo and deprecating the package should not impact that—[previous packages should still be available.](https://docs.npmjs.com/using-deprecated-packages)

DanielRuf:
@DamonHD npmjs (the registry) was meant, not the CLI tool npm (which does not even depend on this package).

Some just conflate and abbreviate npmjs with npm.

MichaelDeBoey:
I indeed meant deprecating the package in the npm registry

DanielRuf:
This will just lead to a simple warning but not block any installs. This is the only thing happening when you use a package, which is marked as deprecated.

MichaelDeBoey:
@DanielRuf That's true, it won't block any installs
But people will get a notification, which makes it clear that they should upgrade.
People looking into the package on npmjs.com will also see the notice and probably choose for HMN instead

